PR #2, opened 2016-07-17
Addded Event and Listener Generation Commands

adelowo:
They follow the same convention as the default commands artisan ships with, e.g the Listener command requires an `--event` option to be present.

### DIFF
--- a/src/Commands/GenerateEventCommand.php
+++ b/src/Commands/GenerateEventCommand.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Nwidart\Modules\Commands;
+
+use Nwidart\Modules\Commands\GeneratorCommand;
+use Nwidart\Modules\Traits\ModuleCommandTrait;
+use Pingpong\Support\Stub;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputArgument;
+
+class GenerateEventCommand extends GeneratorCommand
+{
+    use ModuleCommandTrait;
+
+    protected $argumentName = 'name';
+
+    /**
+     * The console command name.
+     *
+     * @var string
+     */
+    protected $name = 'module:make-event';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Generate a new Event Class for the specified module';
+
+    /**
+     * Get the console command arguments.
+     *
+     * @return array
+     */
+    protected function getArguments()
+    {
+        return [
+            ['name', InputArgument::REQUIRED, 'The name of the command.'],
+            ['module', InputArgument::OPTIONAL, 'The name of module will be used.'],
+        ];
+
+    }
+
+    /**
+     * Get the console command options.
+     *
+     * @return array
+     */
+    protected function getOptions()
+    {
+        return [
+            ['example', null, InputOption::VALUE_OPTIONAL, 'An example option.', null],
+        ];
+    }
+
+    public function getTemplateContents()
+    {
+        $module = $this->laravel['modules']->findOrFail($this->getModuleName());
+
+        return (new Stub('/event.stub', [
+            'NAMESPACE' => $this->getClassNamespace($module) . "\\" . config("modules.paths.generator.event"),
+            "CLASS" => $this->getClass(),
+            'DUMMYNAMESPACE' => $this->laravel->getNamespace() . "Events"
+        ]))->render();
+    }
+
+    public function getDestinationFilePath()
+    {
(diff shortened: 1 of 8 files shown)


src/Commands/GenerateEventCommand.php
```diff
@@ -0,0 +1,97 @@
...
+	protected $description = 'Generate a new Event Class for the specified module';
+
+	/**
+	 * Create a new command instance.
+	 *
+	 * @return void
+	 */
+	public function __construct()
+	{
+		parent::__construct();
```
nWidart: No need for a construct here since it doesn't do anything but calling the parent.

src/Commands/stubs/event.stub
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace $NAMESPACE$;
+
+use $DUMMYNAMESPACE$\Event;
```
nWidart: Why is this line needed ? What's dummy namespace ? Events by default don't really need to extend anything.

src/Commands/stubs/event.stub
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
```
nWidart: You use those 2 classes, but not used in the event class itself.

src/Commands/stubs/listener.stub
```diff
@@ -0,0 +1,31 @@
...
+        //
+    }
+
+    /**
+     * Handle the event.
+     *
+     * @param $EVENTSHORTENEDNAME$ $event
+     * @return void
+     */
+    public function handle($EVENTSHORTENEDNAME$ $event)
```
nWidart: What's the difference between eventshortenedname & eventname ?

nWidart:
Hello,

Thank you for you contribution.

Besides the couple of comments inline, I have those few notes:

- Please follow psr1/2, meaning using spaces instead of tabs.
- Can you add tests for those 2 commands ? You can view an example of a similar test case [here](https://github.com/nWidart/laravel-modules/blob/master/tests/Commands/MakeRequestCommandTest.php)

src/Commands/stubs/listener.stub
adelowo: Eventname is the fully qualified class name (plus namespace).. having to use EVENTSHORTENEDNAME is more like a personal sentiment.. I would fix that though

src/Commands/GenerateEventCommand.php
adelowo: Got it

src/Commands/stubs/event.stub
adelowo: I would remove those

src/Commands/stubs/event.stub
adelowo: The `SerializeModels` class is included in the generated class while the `ShouldBroadCast` is left for the user to implement. Do you want that taken off ?

adelowo:
I had my ide set it (tabs) to 4 spaces, i would just move make it use spaces instead.

src/Commands/stubs/event.stub
nWidart: Thanks 😄

src/Commands/stubs/event.stub
nWidart: Hm I think so yea, and people can implement it as needed. Otherwise the IDE like phpstorm will say it's not used and hint to remove it. :)

src/Commands/stubs/listener.stub
nWidart: Ahh yes I agree, that makes sense :)

nWidart:
Looks like you forgot to register the commands in the CommandServiceProvider class.

Also in a test file you have a double semicolon `;` and the end of a file.

nWidart:
Tests are still failing. I don't think you registered the commands here https://github.com/nWidart/laravel-modules/blob/master/src/Providers/ConsoleServiceProvider.php#L70

nWidart:
Please squash your commits once you have the tests working.

Also please remove inline comments like 

> // Tests fail because of an undefined variable, event

Thanks

adelowo:
Will do as you've asked..

adelowo:
It's been squashed already and I think the new console commands should be added to the `README` file

adelowo:
Any update on this yet

nWidart:
Thank you for the contribution, will update the docs.

nWidart:
Your tests were failing. :/

adelowo:
Thanks a lot for merging :+1:

PR #26, opened 2016-08-13
Adjusted case of module base path to fit folder structure

mash1t:
Folder structure looked strange to me after creating a module, guess that's a typo. According to the readme (sections "Folder Structure" and "Autoloading") modules it is meant to be lowercase. On a case sensitive file system, this is crucial, because namespaces can't be autoloaded.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -62,7 +62,7 @@ return [
         |
         */
 
-        'modules' => base_path('Modules'),
+        'modules' => base_path('modules'),
         /*
         |--------------------------------------------------------------------------
         | Modules assets path


nWidart:
Hello,

Thanks but this is intended, since https://github.com/nWidart/laravel-modules/commit/b29bcb26c1470c5663a1138f83acb215accf5b84 .

mash1t:
May i ask why the lowercase versions are still in the readme? As i pointed out, this way the autoloading isn't working properly

nWidart:
That's my bad, I must have forgotten to update the readme. 😄

PR #32, opened 2016-08-18
Added force option to module:seed command

gorog:
Right now, it's impossible to use the mode:seed command without interactions in a production environment.
The changes make it possible. :)


nWidart:
Thank you, very helpful indeed! 😄

PR #40, opened 2016-08-30
Remove unnecessary loops in register and provides

mikemand:
`$this->commands()` will accept an array and do the loop itself.

`provides()` returns an array, which the commands are already in.


nWidart:
Thanks!

PR #80, opened 2016-10-23
Improve config publishing

gjmh-unikent:
I had some issues publishing configs from a module using `php artisan module:publish-config` since i was using a custom namespace, i.e. not "Modules\...".
Additionally the service provider stub that is generated does not tag config files correctly for this command.

### DIFF
--- a/src/Commands/PublishConfigurationCommand.php
+++ b/src/Commands/PublishConfigurationCommand.php
@@ -2,6 +2,7 @@
 
 namespace Nwidart\Modules\Commands;
 
+use Module;
 use Illuminate\Console\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


nWidart:
Hello,

Thanks for the contribution.

Not sure I understand the need for this.
Publishing configuration files works if you have your module registering them.

This was the behaviour that was intentionally removed in order to move that responsability to the module.

gjmh-unikent:
I am creating some modules that will be pulled in via composer i want them to use my own vendor namespace.
I have changed the `namespace` config option and generated a new module.
The `module:publish-config` command always assumes a modules service provider is in the "Modules" namespace and thus is does not publish the config from the new module.
This change allows a module to use any namespace, and have the config publishing still work.

In addition the service provider stub does not set the `config` tag when registering the generated config file meaning it will not be published even if not using a custom namespace. you will receive a `Nothing to publish for tag [config].` message.

gjmh-unikent:
Realised i had left my module key hardcoded, have switched this to use the functions parameter correctly, perhaps this is what had you confused.

nWidart:
If the namespace was the issue, it could just use the configuration option on the namespace (if none present create one).

gjmh-unikent:
I considered that, however that would just present the same issue if using modules with different namespaces, better to let each module provide its fully qualified service provider(s).

That way i could composer in modules from any number of different vendors, and load in something locally using the default Modules namespace or a custom one, and publishing would work for all of them.

argentinaluiz:
Will this pull request be approved?

armababy:
This currently would break default behaviour, as for example current modules host number of providers for which you would not necessarily want to publish config and can actually accidentally overwrite them if force flag used.

gjmh-unikent:
I can see that could be undesirable, perhaps an addition to the module.json for "publishes" to allow you to explicitly list the service providers (with fully qualified class names)  that you wish to be published via the artisan commands.

nWidart:
Closing this as it's not needed currently. 😄

PR #88, opened 2016-11-11
fix bug: module:migrate-reset and module:migrate-rollback don't use t…

HuNanZai:
hi there,
i look carefully into the code and i found:
1. this problem only locates in  migrate-reset and migrate-rollback command
2. laravel itself provide db connection choose function
so i fix this bug by:
1. add a property(and a set method) in the Migrator class
2. use the connection in the table() function
3. set the connection in the two command's relate function
hope this will work!

### DIFF
--- a/src/Commands/MigrateResetCommand.php
+++ b/src/Commands/MigrateResetCommand.php
@@ -61,6 +61,12 @@ class MigrateResetCommand extends Command
 
         $migrator = new Migrator($module);
 
+        $database = $this->option('database');
+
+        if (!empty($database)) {
+            $migrator->setDatabase($database);
+        }
+
         $migrated = $migrator->reset();
 
         if (count($migrated)) {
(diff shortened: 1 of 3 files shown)


nWidart:
This looks good, thanks! 👍 

One small question, if no connection is specified, it will use `->connection('')`, with an empty string as argument, will that use the default connection internally ?

Thanks,

HuNanZai:
it won't.
the ` ? :` expression will take the empty string '' as a no condition.
you can write a simple test:  `php -r "echo '' ? 'yes' : 'no';"`, it will tell.

nWidart:
Sorry for late reply, thanks for the contribution! 👍

HuNanZai:
it's necessary to let u know that there is still a small "problem", sort of...
because when you use `migrate-rollback` or `migrate-reset` command, the `--database` option only affects the build process of the Migrator Class.
so u still have to specific the db connection in the migration files of your module to make it work.
Here are the related code in the Migrator.php 
```
public function down($migration)
{
    $this->resolve($migration)->down();    //don't specific the connection
}
```
i don't know whether this is a "problem".
As far as i know, laravel migration class doesn't have the modify access method,
So in my opinion, it's better to write down this tip in the README.md 😄

PR #91, opened 2016-11-17
Fixed Module::asset();

henrydinhh:
Remove a frist slash.

Before fixed:

```
Module::asset('system:css/app.css');
// http://localhost:8000/\modules/system/css/app.css
```

After fixed:

```
Module::asset('system:css/app.css');
// http://localhost:8000/modules/system/css/app.css
```


nWidart:
Thanks 👍

PR #94, opened 2016-11-23
Update controller.stub

mjmarianetti:
Add the show method to the controller stub to show a particular resource.


nWidart:
Hi, thanks for your contribution.

Can you please adapt the failing text according to the change you made ?

Issue #96, opened 2016-11-28
php artisan module:migrate doesn't check time stambs 


ercanertan:
I was trying to run **php artisan module:migrate** but there is a issue about migration.

I am assuming migration running by alphabetically. Is that right?

armababy:
Migrations should be sorted [over here](https://github.com/nWidart/laravel-modules/blob/master/src/Migrations/Migrator.php#L103).
Can you give some filename examples, expected migration order and in what order actually they were run by laravel-modules.

ercanertan:
Example here

2016_11_28_144737_create_services_table.php
2016_11_28_155717_create_services_translations_table.php

2016_11_28_104907_create_service_blocks_table.php
2016_11_28_105059_create_service_blocks_translations_table.php

My expectation was following
1.Migrate "service_blocks"
2.Migrate "services"

But migration was first creating "services" which is why I was having Foreign key error

nWidart:
Look at your timestamps, they're not in order. That's why `2016_11_28_155717_create_services_translations_table` was sorted first.

nWidart:
I cannot reproduce this as migrations have always been sorted before.

armababy:
Can you give us environment on which this behaviour occurred? Mainly OS, if any virtualization (host, guest) etc.
Thanks.

ercanertan:
Thanks for the quick respond, probably I will create migration files again.

ercanertan:
I am using latest Mac OS X, Valet, Mysql 
thanks

armababy:
So I did a little bit of testing and pretty much could not get sort wrong :(

So the command `module:migrate` actually does not call `getMigrations` method in laravel-modules `Migrator` class as I initially thought.

What actually happens is `Migrator` class only returns module migrations folder path and then the path is passed to Laravel's own `migrate` command. And the getting and sorting of files happen there.
And in this instance laravel-modules does no sorting whatsoever.

In my humble tests, i always got order right as you expected.

Internally laravel uses `sortBy` on a collection of files found and in turn `sortBy` uses (in this case) php function `asort` (low to high) sorting function.

It would be really great if you could try on some other system or setup. So we could pinpoint this problem that might be happening.

![migrations-test](https://cloud.githubusercontent.com/assets/219556/20734854/ea13e344-b6a4-11e6-805f-cae5c2fdad0c.png)

ercanertan:
@armababy thank you for your time. I will do the test however

Could you pls try from different modules.

[...]

PR #98, opened 2016-11-30
Respect module order in reverse migrations

armababy:
Based on #96, commands like `module:migrate-rollback` and `module:migrate-reset` does not respect order parameter of modules as `module:migrate` does.

Also instead of all modules it will run only on enabled ones.

Added `--direction`/`--d` command to both commands defaulting to `desc` (all enabled modules in reverse order).


nWidart:
Awesome, thanks! 👍

ercanertan:
Thanks

nWidart:
Can you confirm on dev-master your issue is fixed ? If so, I can tag the release. 😄

PR #116, opened 2017-01-10
Add Macroable trait to module repository

gmox:
At my job, we use this library a lot in a very large project, and if there's one thing that would be extremely useful in creating custom functionality that extends your library, it's being able to define custom macros onto the repository.

This PR adds two tests for the Macroable trait, and adds the Macroable trait to the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,6 +541,19 @@ Update dependencies for the specified module.
 Module::update('hello');
 ```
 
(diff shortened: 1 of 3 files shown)


nWidart:
Hello & thank you for your contribution!

This looks very nice & handy, thanks!

PR #117, opened 2017-01-11
Module requirements

gmox:
One thing we do often at the company I work for is have modules require other modules. We used to go by "order" but that only goes so far: explicit requirements are extremely useful. The ability to say "I need module X, and X requires functionality from Y" has been a reoccurring theme. Therefore, I submit this PR.


nWidart:
Hello,

This looks good. 
I'm assuming you have something on your end that uses that method to require the required modules ?

gmox:
At the moment, yes. I was debating about adding it into `Module::register()` here, but some people may want to have different requiring procedures (on a SAAS platform, for instance, you may want to check that someone's plan allows a certain module before enabling it). But with that being said, listening for the `module.enabling` would allow a great range of functionality with this.

nWidart:
I see, thanks. 👍

PR #119, opened 2017-01-12
Change default namespace for Mailables to match Laravel default

mikemand:
I would have sworn they had Mailables in an `Email` directory when they first came out, but maybe I was just projecting...


mikemand:
Closing. See AsgardCMS Slack for reasoning.

PR #127, opened 2017-01-19
Fixed typo in README.MD

ecodrutz:
It seems the description and the code for Enabling/Disabling a module was wrong. I believe the correct ones are:
Disable the specified module.
 $module->disable();
  Enable the specified module.
 $module->enable();


nWidart:
Good catch! Thanks! 👍

PR #152, opened 2017-02-26
Add checking for failure to parse module JSON

gmox:
There's not much worse than spending a good chunk of time bug hunting because you left a trailing comma at the end of your module.json file.

This will check if there was an error and throw an exception.


nWidart:
Nice this looks useful! 👍

Issue #160, opened 2017-03-04
ListCommand outputs StudlyName instead of actual name


tentus:
The module:list command outputs the StudlyCase version of each module name, which is problematic if you're trying to use the name in conjunction with other commands, which need the actual name to work (eg module:enable or module:disable).

It's also rendered somewhat redundant by the Path column, which would under normal circumstances include a StudlyCase version of the module name. (Running module:make with the name "an Example Module" would create a path with AnExampleModule at the end.)

Lastly, outputting the module name directly doesn't represent any loss in data: anyone relying on that column to be StudlyCase can easily apply an equalivalent method on their end, while the reverse is not true. Module names that are acronyms are particularly vulnerable to this.

gmox:
It looks like it was the leftover from pingpong modules, which brings me to my question: does laravel-modules need to keep doing that?

I'm just taking a guess, but it looks like it was for filesystem reason. However, there's better ways to solve that problem than have many ways to refer to a module by name.

gmox:
@nWidart maybe you have some input?

nWidart:
Sorry for the late reply.

I agree that indeed the output shouldn't be formatted, but just use the name instead.

nWidart:
Merged & tagged. Thanks! 👍

PR #164, opened 2017-03-13
Change list command to give the name of the module, not the studly cased form

gmox:
From the discussion of https://github.com/nWidart/laravel-modules/issues/160


nWidart:
Thank you! 👍

PR #167, opened 2017-03-15
Making the $migrations parameter optional on getLastBatchNumber()

joshbrw:
When manually running the `->log()` method externally (public ACL) it fails as it tries to call the `getNextBatchNumber()` without a parameter.


nWidart:
Thanks! 👍

PR #168, opened 2017-03-16
Add a subpath option to migrate command

Fagl4:
Adds a subpath option to the migrate command so the user can structure different migrations per folders


nWidart:

I thought this was PR'ed by someone else already, but must have been a discussion on an issue. 😄

PR #194, opened 2017-04-03
Add Phpstorm unversioned directory to .gitignore

poppabear8883:
Just accept or add it. Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 /fixture/vendor
 build
 .php_cs.cache
+.idea/


mikemand:
This would be something to put in your personal global `.gitignore`. See [here](https://thomashunter.name/blog/global-gitignore-vs-repository-gitignore/) for more information.

poppabear8883:
Sure that's one way to do it :wink:

nWidart:
Indeed this is something related to your personal environment (even if we also use phpstorm), so that should be in your global gitignore file.

Thanks for the contribution though! 👍

PR #199, opened 2017-04-12
Changing path to /Listeners/ as this is the path Laravel now recommends

joshbrw:


nWidart:
This is something possible via configuration. I'm personally not a fan of this change.

joshbrw:
This is the standard directory that is advised by Laravel. Surely it's best to follow the best practices dictated by the framework the package is for?

https://laravel.com/docs/5.4/events#defining-listeners

nWidart:
Why would we need to follow every guideline proposed by the framework ? 

The change is easy to make via configuration if it's really something you need.

JulienTant:
You may reconsider your position, Nicolas. 
Being consistant with the framework's standard is very important, so the developper experience is better.
And, it personally makes more sense to me to have Listeners (or Handlers) in their own "root" folders. Not in some subfolder of events.

thetar:
I must agree with @joshbrw and @JulienTant. The only reason it is located where it is currently is to follow Laravel standards, and when those change, we should adapt. Personal opinion aside, default behavior should always follow the tools you choose.

nWidart:
Implemented since https://github.com/nWidart/laravel-modules/commit/387629605132417d32df1b149d3f9e54ebf5bc33

PR #200, opened 2017-04-17
Moved the registering of Modules to register()

thetar:
For your consideration:

Module service providers should be allows to register binding via register() at the same time as all over service providers. This doesn't change the timing of the boot() functionality.

The use case I ran into on this is the registering order of the bindings, I have service providers for different traffic sources (base on the url being used to access the site) These service providers need to be able to register bindings last in order to override the defaults. So that ABC.com binds Products to ABCProducts for that sites product behavior differences. With modules not binding until the boot phase, it makes modules always override any other service providers register() method.

### DIFF
--- a/src/LaravelModulesServiceProvider.php
+++ b/src/LaravelModulesServiceProvider.php
@@ -23,8 +23,6 @@ class LaravelModulesServiceProvider extends ServiceProvider
     public function boot()
     {
         $this->registerNamespaces();
-
-        $this->registerModules();
     }
 
     /**
@@ -43,6 +41,7 @@ class LaravelModulesServiceProvider extends ServiceProvider
         $this->registerServices();
         $this->setupStubPath();
         $this->registerProviders();
+        $this->registerModules();
     }
 
     /**


thetar:
Any thoughts on merging this in? If not, I'll have to fork off in my current project so I'm just wondering before going through the effort.

nWidart:
Hello,

I'm very sorry for the lack of response! I haven't merged this yet because I suspect this is going to be a breaking change. For instance in AsgardCMS which relies on the placement of this method. 

I just tried it out, and indeed it breaks the applications. It seems easy and not breaking change but the order of loading things is pretty major. 😄 

I'm thinking of a solution for you without having to fork the package.. You might be able to create your own `LaravelModuleServiceProvider` class, change the order there. You can then  load your own custom SP instead of the one from the package, every should still work as you register the same logic.

Let me know what you think of this possible solution.

thetar:
Figured it was something that would have to go through some testing. It seems to me that it is a question between what is correct technically and following the Laravel bootstrap process I think that the change makes it the most sound on a technical level to the purity of the Laravel world. There are things that need to be registered prior to anyone else booting, modules shouldn't be an exception to the rule.

What breaks in your CMS? Is it just an oversight on placement, and you could lock version compatibility and move forward.

If it makes more since to hold off until a major version, I can wait, easy enough to fork. I don't have an issue forking if I must, the CMS I'm working on will require a lot of control in logic at a high/entry point of the boot process that others might not require. I have always been curious on just how much of the actual logic I need outside of the simply directory mapping and it would give me a chance to assess that as well.

nWidart:
No, I'm not going to lock the whole cms development just because of one change. 😄 

It's very easy for you to use the solution above. Ie registering your own service provider that does that extends this one, with the move of methods.

Error is "no class cache", why it's in boot to have that cache binding done.

thetar:
Of course not, maybe you should add a test that I would use to test different methods or correcting the core issue without affecting this edge case?

The above bandaid wouldn't work out well for the future of my projects. I have forked for now, again, no rush, I'll just rework and get it off dev-master if needed by the time I need these to launch stable.

Thanks for clarification.

nWidart:
Don't know what you're looking for with a test. If a module depends on a core laravel dependency the framework needs to be booted.

I think creating a service provider is faster than you having to maintain a fork, out of experience. This is also one of the use cases of service providers in laravel. Creating 1 class, doesn't require any maintenance on your part, compared to a fork for example.

thetar:
As you know, there is a sequence in the bootstrap that is followed and the sequence of how service providers are loaded into the `config/app.php` determines the order of which they register and boot. 

However, this package does not register it's service providers until the `boot()` method. This means nothing can load behind it unless it also **registers** its providers in the `boot()` method, which is not the correct location and breaks things unintentionally like you are seeing with AsgardCMS.

The system starts to boot before the modules even start to register (almost all the way at the end of the system boot, which will cause a lot of side effects.

Instead of:
1. register
1. register
1. register
1. boot
1. boot
1. boot

You get:
1. register
1. boot
1. boot
1. register
1. register
1. boot

I will install AsgardCMS tonight and see if the fix is what I suspect it is, which is a simple sequence of events issue. I'll comment back with some relative examples to AsgardCMS.

I'm not pushing for this to be merged in, it's your library, but I feel like it should play nice/fair to it's environment/intent. I know you took it over from pingpong so I figured since I ran into the one use case that this would matter for that I would point it out. (btw, thank you for stepping up and doing that when pingpong fell off the planet:))

thetar:
I see the issue, you are making the common mistake of using a service provided by a different service in the register method. Here is a [very good explanation](https://laracasts.com/discuss/channels/general-discussion/difference-between-boot-and-register-method) of the issue.

The main reason you reserve binding's and other logic to the boot() method is because you can ensure your services (like cache) has already been registered and ready to use. Same for registering, you don't want to register inside the boot() method because other service providers that rely on those could boot before they are registered and blow up. 

I went through you AsgardCMS code, but I didn't have enough time to change all the service providers to get it up and running. If you google your error you will get a lot of resources talking about the same bootstrap process. Like [this one](http://stackoverflow.com/a/36772858/2330224)

nWidart:
Sorry for the late reply.
I'm away of the difference between the 2 methods, thanks.

I'll close this for now as it's a breaking change, and very easy to customise on your end without having to fork.

thetar:
Clearly not because otherwise it is not a "breaking change".. :) No one that used the methods as intended would be effected and wouldn't be objected to side effects of the breaking of the boot chain.

nWidart:
Now if you've read that book, you would understand what service providers are, and why they're useful.

I've setup a quick test project for you to look at how easy it was to overwrite this requested functionality.
https://github.com/nWidart/laravel-modules-custom-service-provider

Take a look at this : https://github.com/nWidart/laravel-modules-custom-service-provider/blob/master/app/Providers/CustomLaravelModulesServiceProvider.php

Which is used here https://github.com/nWidart/laravel-modules-custom-service-provider/blob/master/config/app.php#L170

thetar:
"Is in my opinion incorrect" ..it's not an opinion, it is how your modules are registered, after others start booting.. As I said, nothing should boot before EVERYTHING is registered. This library, as is, allows everything before LaravelModulesServiceProvider to boot before it starts to register modules service providers.

Like I said, I already forked, so correct it or don't. Just don't attack me about the issue, because it is an issue and if the register and boot methods are use as intended (as they should be), there is no side effect to the change. and would correct the issue that is in fact an oversight in how modules are handled.

Adding an App level service provider because the modules load incorrectly is an insane hack that completely defeats the reason you would want modules in the first place.

nWidart:
"Insane hack" ? That's why service providers exists, thus not a hack, but intended use. 
A fork a the insane hack (keeping it up to date, etc.). 

Yes it's an opinion. The boot methods are called last, ie you can't have a register, after a boot.

thetar:
> The boot methods are called last, ie you can't have a register, after a boot.

My point exactly, with your library you can. even with an app level service provider to fix the oversight, your modules are not being treated properly and you can run into side effects with it. I guess you are missing the issue all together.

"My fork" is is not that one on this account. :)

getriebesand:
I run into the same problem now. The register methods schould run before boot methods of all service providers. Not only before boot methods of module service providers.
Current implementation does not follow the livecycle of service providers [https://laravel.com/docs/5.6/lifecycle](https://laravel.com/docs/5.6/lifecycle)

Please fix that because thats a laravel lifecycle breaking bug

PR #205, opened 2017-04-22
added 5.5.* to required list of Laravel versions

dydx:
Tried using this with `5.5-*` and found it wouldn't install. Added `5.5.*` to the required versions list and everything worked fine with using my forked branch as the Composer repository / source.


nWidart:
Thanks! 👍

PR #214, opened 2017-05-02
Update Repository.php

FaZeRs:
Get collections by status

### DIFF
--- a/src/Repository.php
+++ b/src/Repository.php
@@ -396,7 +396,7 @@ class Repository implements RepositoryInterface, Countable
      *
      * @return Collection
      */
-    public function collections($status)
+    public function collections($status = 1)
     {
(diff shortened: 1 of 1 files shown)


mikemand:
You should provide a sane default for the status.

src/Repository.php
```diff
@@ -392,11 +392,13 @@ public function findOrFail($name)
     /**
      * Get all modules as laravel collection instance.
      *
+     * @param $status
+     *
      * @return Collection
      */
-    public function collections()
+    public function collections($status)
```
mikemand: You should provide a sane default for the status.

nWidart:
Thank you for the contribution 👍 
Indeed having the default avoids a breaking change.

PR #218, opened 2017-05-09
Add Macroable trait to Module

tentus:
In the project I work on we have some additional organization applied to our modules, which would be more elegant to work with if we could macro into the modules themselves some awareness of that structure. The Macroable trait is a simple solution to that problem.


nWidart:
Very nice addition, thank you! 👍

PR #220, opened 2017-05-10
Minor inspection cleanup

tentus:
I did a quick pass on a set of inspection results, mostly cleaning up places where a return value was documented but never actually happened, etc. I didn't try to be super thorough, just trimming down some of the low-hanging fruit that IDEs like PHPStorm will complain about.


nWidart:
Thank you very much! 👍

Issue #202, opened 2017-04-18
Impossible to publish configs due hardcoded value


webmake:
Hi, there is no possibility to publish configs, when having custom namespaced service provider, due hardcoded `Modules` word:
```
    private function getServiceProviderForModule($module)
    {
        $studlyName = studly_case($module);

        return "Modules\\$studlyName\\Providers\\{$studlyName}ServiceProvider";
    }
```
Our case, when we have under company namespace.
seems config has no effect:
```
    /*
    |--------------------------------------------------------------------------
    | Module Namespace
    |--------------------------------------------------------------------------
    |
    | Default module namespace.
    |
    */

    'namespace' => 'Company',
```

for example:
`module:publish-config`
generates following command:
`vendor:publish --provider="Modules\\Test\\Providers\\TestServiceProvider"`

nWidart:
Good catch, it should indeed use that namespace config value. Will fix when possible.

webmake:
Also notable, there is trick with vendors, we also use vendor scan, so namespace can differ, I'm not sure how to bypass this using this module.. 
```
    'scan' => [
        'enabled' => true,
        'paths' => [
            base_path('vendor/OtherCompany/*/src'),
        ],
    ],
```

nWidart:
This is now fixed on master branch, tagging as soon as Travis is green.

For your second issue however it is currently not supported to have multiple namespaces for different modules.

webmake:
Thanks.

Yeah, for multiple namespaces solution would a little bit harder. Seems enough to add into `Module` class namespace, that
```
        foreach ($this->laravel['modules']->enabled() as $module) {
            $this->publishConfiguration($module->getNamespace());
        }
```
would take it, but 
```
        if ($module = $this->argument('module')) {
            $this->publishConfiguration($module);

            return;
        }
```
 would need to load module again somehow and reparse..

nWidart:
I'll close this now as the main issue was resolved.

The other issue seems more like an edge case, as modules should be in the same namespace ideally. 
However, if you have a working setup of it working, feel free to send a pull request for me to look at.  👍

PR #221, opened 2017-05-10
Take first from providers and reuse it

webmake:
Take first from providers and reuse it, instead of generating misleading paths when having multiple namespaced vendors

fixing https://github.com/nWidart/laravel-modules/issues/202

### DIFF
--- a/src/Commands/PublishConfigurationCommand.php
+++ b/src/Commands/PublishConfigurationCommand.php
@@ -44,10 +44,9 @@ class PublishConfigurationCommand extends Command
      */
     private function getServiceProviderForModule($module)
     {
-        foreach ($this->laravel['modules']->enabled() as $moduleObject) {
-            if ($module == $moduleObject->getName() && count($moduleObject->providers) == 1) {
-                return $moduleObject->providers[0];
-            }
+        $concreteModule = $this->laravel['modules']->find($module);
+        if ($concreteModule && count($concreteModule->providers) == 1) {
+            return $concreteModule->providers[0];
         }
         $namespace = $this->laravel['config']->get('modules.namespace');
         $studlyName = studly_case($module);
--- a/src/Commands/PublishConfigurationCommand.php
+++ b/src/Commands/PublishConfigurationCommand.php
@@ -44,6 +44,11 @@ class PublishConfigurationCommand extends Command
      */
     private function getServiceProviderForModule($module)
     {
+        foreach ($this->laravel['modules']->enabled() as $moduleObject) {
+            if ($module == $moduleObject->getName() && count($moduleObject->providers) == 1) {
+                return $moduleObject->providers[0];
+            }
+        }
         $namespace = $this->laravel['config']->get('modules.namespace');
         $studlyName = studly_case($module);
 
--- a/tests/Commands/PublishConfigurationCommandTest.php
+++ b/tests/Commands/PublishConfigurationCommandTest.php
@@ -39,7 +39,7 @@ class PublishConfigurationCommandTest extends BaseTestCase
             'command' => 'vendor:publish',
         ];
 
-        $this->mockLaravelPublish($arguments);
+        $this->mockLaravelPublish([$arguments]);
 
         $this->artisan('module:publish-config', ['module' => 'Blog']);
     }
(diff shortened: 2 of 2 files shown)


nWidart:

Could you add test(s) for this feature to make sure it keeps working in the future please ? 

Thanks

webmake:
Tomorrow I will try to update this PR :)

webmake:
Seems now not failing

nWidart:
Hello,

I'm trying to understand how this fixes your issue of multiple namespaces for multiple modules.
How does it fix it ?
``` php
$concreteModule = $this->laravel['modules']->find($module);
        if ($concreteModule && count($concreteModule->providers) == 1) {
            return $concreteModule->providers[0];
        }
```
From what I see is you search for the module, if that module exists and it has a service provider, take the first service provider.
This seems pretty risky & a breaking change for people that have a different order.

webmake:
It fixes config publishing, because having multiple namespaces, there is no other option to retrieve namespace associated with module, except via module.json itself..
It doesn't break nothing (or at least I don't know such case). That method is called only on publishing configs via artisan, and scans module.json in same manner like your service provider acts while registering modules, same order, and additional guard (like you said) prevents from loading multiple service providers. So it must exist only one that publish config, in other case with multiple declarations in module.json - would work as previous by default.

webmake:
You can take a look at https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/json.stub#L8
this line `count($concreteModule->providers) == 1` works only just then 1 registered

nWidart:
I see, but since it takes the first provider listed, what if the first provider listed, is one of a random package ?

webmake:
Then project shouldn't work at all I guess, not even config publishing. Because modules loading works according that json.

nWidart:
Hm don't think so as it uses the configuration value.

webmake:
Let's get this clear from very beginning :) This functionality was mentioned in previous [discussion](https://github.com/nWidart/laravel-modules/issues/202)
I modified private method `getServiceProviderForModule`, and this is used **only** in `publishConfiguration`, when running from artisan `module:publish-config`, so it doesn't break anything further.

This PR extends with ability to publish **not only** under module [directory](https://github.com/nWidart/laravel-modules/blob/master/config/config.php#L14), ex.: having `'namespace' => 'ModuleProjectNamespace'` original method would make **ModuleProjectNamespace**\\**ModuleName**\\Providers\\**ModuleName**ServiceProvider.
Once enabled
[scan](https://github.com/nWidart/laravel-modules/blob/master/config/config.php#L129), it didn't build proper namespace. Let's say having secondary [path](https://github.com/nWidart/laravel-modules/blob/master/config/config.php#L131). As workaround we used default laravel command:

vendor:publish --provider="**Vendor**\\**Package**\\Providers\\**Package**ServiceProvider"
in compare with current module functionality:
vendor:publish --provider="**ModuleProjectNamespace**\\**Package**\\Providers\\**Package**ServiceProvider")

With this functionality we bring possibility again to publish module's config, no matter under what namespace it exists .
BTW, I saw [PR](https://github.com/nWidart/laravel-modules/pull/80) that works pretty similar like this one, but I left default namespace building if there exists more providers in module.json file as uncertain behaviour, like you told order or smth other.  Again, no breaks.

There wasn't fully exploit the potential for `providers` key previously, because only module registering was using it `registerProviders`, so I still don't see any problems reusing it.

Actually tests pretty well reflects all cases:
With 1 module named `Blog`:
* having 1 provider (`Vendor1/PackageNamespace/ServiceProvider`) in json - publishing Blog, it will publish `Vendor1/PackageNamespace/ServiceProvider`
* having more than 1 (`Vendor1/PackageNamespace/ServiceProvider`, `Vendor2/PackageNamespace/ServiceProvider`) in json - publishing Blog, it will publish `Modules\\Blog\\Providers\\BlogServiceProvider` (default)

With 2 modules `Name1Module`, `Name2Module`, having respectively `Vendor1/PackageNamespace/ServiceProvider`, `Vendor2/PackageNamespace/ServiceProvider` - and  publishing all modules without module parameter
* it will publish  `Vendor1/PackageNamespace/ServiceProvider` and `Vendor2/PackageNamespace/ServiceProvider` accordingly

nWidart:
Thanks for taking the time to explain.

I did understand it's only for the publishing of configuration. I also understand it's to allow multiple module namespaces.

And I still stand by the point that it will fail, if you set an external service provider first, in the `providers` key.

Take this providers array:

``` json
"providers": [
    "FloatingPoint\\Stylist\\StylistServiceProvider",
    "Modules\\Core\\Providers\\CoreServiceProvider",
    "Modules\\Core\\Providers\\SidebarServiceProvider",
    "Modules\\Core\\Providers\\AssetServiceProvider",
];
```

This will use the FloatingPoint namespace correct ?

webmake:
It won't execute at all, because, count is 4

nWidart:
Ok, and

``` json
"providers": [
    "FloatingPoint\\Stylist\\StylistServiceProvider",
];
```

webmake:
Yes, testing with integrational testing here would be enormous preparation, to create separate folders for each test cases, and modules with in it :smile:  so I mock everything what is out of module scope

If I get you right, you mean that declaration for non existing.
So generated command from `module:publish-config` would be `vendor:publish --tag=config --provider="FloatingPoint\\Stylist\\StylistServiceProvider"`, despite that module name itself would be different than **Stylist**.
Running any not existing provider (simplified to **asdfg**)
```
$ php artisan vendor:publish --tag=config --provider=asdfg
Publishing complete.
```
nothing happens, no errors on non existing provider.

Also there should be such case, because project wouldn't work at all, it gives an error:
`Class 'FloatingPoint\Stylist\StylistServiceProvider' not found`
```
at Module->registerProviders() in Module.php line 231
at Module->register() in Repository.php line 296
```
So it happens on registering.

webmake:
If we switch context, and module `A` have 
```
"providers": [
    "Common\\B\\BServiceProvider",
];
```
and module `B` have
```
"providers": [
    "Common\\A\\AServiceProvider",
];
```
then project works, but publishing `module:publish-config A` configs it generates `BServiceProvider`. This is expected workflow, because registering providers, with module `A` is associated `BServiceProvider`. Or do you want to publish according module A
`vendor:publish --tag=config --provider="Common\\A\\AServiceProvider"`, which is belonging for `B`?

nWidart:
I'm not saying it doesn't work. I'm just pointing out that it assumes the first provider is one from the module, which doesn't have to be the case.

unxp:
Hey. I'll try to step in and clear things up a bit :)

I don't know about the validity of the PR, but I do know the problem itself. So I'll try to explain it and maybe you will know the solution.

We are currently developing our modules as packages and import them to our projects via composer. Every package has their own `module.json` file and acts as `nWidart` module. To make our packages to be loaded, we publish `module.php` config and enable scanning like so:

```php
'scan' => [
    'enabled' => true,
    'paths' => [
        base_path('vendor/our-company/*/src'),
    ],
],
```

Every module is being loaded and everything works fine. However, `php artisan module:publish-config SomeModule` can not be used. It finds the module, just no files are published. Keep in mind that our packages are under different namespace than the one defined here:

```php
/*
|--------------------------------------------------------------------------
| Module Namespace
|--------------------------------------------------------------------------
|
| Default module namespace.
|
*/

'namespace' => 'Modules',
```

Their namespace is something like this: `OurCompany\SomeModule`.
Now, if we use `php artisan vendor:publish --tag=config --provider=OurCompany\SomeModule\Providers\SomeModuleServiceProvider` it works and files are being published.

Maybe you could identify where the problem lies and fix it or just guide us on how to fix it and we will update the PR :)

Thank you in advance.

unxp:
@nWidart ping :)

nWidart:
Hello & sorry for the late reply.

I'm not sure why you felt the need to re-explain the issue.
I don't think this is the best solution to the issue for the reasons I stated before.

webmake:
So how you suggest to use vendor scan if that functionality doesn't work?

nWidart:
People are using the vendor scan so it does work.

Also take a look at https://github.com/joshbrw/laravel-module-installer which might interest you.

PR #226, opened 2017-05-18
Update GenerateNotificationCommand.php

manojsb:
There was no description for this command in terminal. So just added that.


nWidart:
Awesome, thanks for the contribution! 👍

PR #244, opened 2017-06-12
Change base_path('resources/...') to resource_path('...')

pixelarbeit:
This won't break functionality if the folder structure of laravel was changed and the resource path was set correctly.


nWidart:
This is nice, thanks!

PR #245, opened 2017-06-12
Some features

wffranco:
* Module::isActive('module') let you validate if a module exists and is active but without generating error.

* Routes folder, configured as in Laravel 5.3+ (and according to the standard, the routes should be loaded in the service provider, so the start.php file is still present but empty).

* No Namespace folders (like config & routes) must be in lowercase.

### DIFF
--- a/src/Repository.php
+++ b/src/Repository.php
@@ -553,6 +553,18 @@ class Repository implements RepositoryInterface, Countable
     }
 
     /**
+     * Determine if the given module exist and is activated.
+     *
+     * @param string $name
+     *
+     * @return bool
+     */
+    public function isActive($name)
+    {
+        return ($module = $this->find($name)) && $module->active();
+    }
+
+    /**
      * Enabling a specific module.
      *
      * @param string $name
--- a/config/config.php
+++ b/config/config.php
@@ -28,14 +28,18 @@ return [
         'files' => [
             'start' => 'start.php',
             'routes' => 'Http/routes.php',
+            'routes/api' => 'routes/api.php',
+            'routes/web' => 'routes/web.php',
             'views/index' => 'Resources/views/index.blade.php',
             'views/master' => 'Resources/views/layouts/master.blade.php',
             'scaffold/config' => 'Config/config.php',
             'composer' => 'composer.json',
         ],
         'replacements' => [
-            'start' => ['LOWER_NAME'],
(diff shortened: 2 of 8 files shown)


src/Repository.php
```diff
@@ -553,6 +553,18 @@ public function notActive($name)
...
     /**
+     * Determine if the given module exist and is activated.
+     *
+     * @param string $name
+     *
+     * @return bool
+     */
+    public function isActive($name)
+    {
+        return ($module = $this->find($name)) && $module->active();
```
nWidart: The `active` method does this already.

nWidart:
Hello and thanks for the contribution.

However please consider sending one PR per feature/change you make.

src/Repository.php
wffranco: Yes but the actual "active" method throw an error if the module is not present.

The new method let you validate if the module exist and is active.

Example: I need to create a menu based on 3 optional modules. So if a partner install just 2 modules, the menu throw an error with the old active method. The new one just return true or false.

wffranco:
Except for the "isActive" method, all the other files are related, and should be in the same pull request.

src/Repository.php
wffranco: I preserve the original functionality in case somebody else need it that way.

nWidart: It will throw a `ModuleNotFoundException` exception, which you can catch to build your menu.

nWidart:
Sorry but after further review I can't accept this PR. Routes are loaded in a RouterServiceProvider, which can be generated with this package. This is the usual recommended way instead of via `start.php`.

This also gives the package consumer the flexibility of having the routes file at any desired location.
This is the stub file for the routeSP https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/route-provider.stub
Which is called by the `module:route-provider` command.

src/Repository.php
wffranco: I understand your point, but the main purpose of laravel is easy coding.
So if you need a fast validation, this:
```php
if (Module::isActive('module') {
    // do something
}
```
is faster that:
```php
try {
    if (Module::active('module') {
        // do something
    }
} catch (ModuleNotFoundException $e) {
}
```
that for every module I use.

That behavior make me add this extra code in every module for exceptions that I don't need to catch, I just want to ignore some code if the module don't exist or is disabled.

wffranco:
When I create a new module this don't have the route provider... Is a new feature or somebody disable it?

PR #249, opened 2017-06-19
Add support for generic git host

tomgrv:
1/ Add specific support for "gitlab" repo type during install
2/ Add support for generic git repo type during install

### DIFF
--- a/src/Process/Installer.php
+++ b/src/Process/Installer.php
@@ -150,8 +150,12 @@ class Installer
      */
     public function getProcess()
     {
-        if ($this->type && $this->tree) {
-            return $this->installViaSubtree();
+        if ($this->type) {
+            if ($this->tree) {
+                return $this->installViaSubtree();
+            }
+
+            return $this->installViaGit();
         }
 
         return $this->installViaComposer();
--- a/src/Process/Installer.php
+++ b/src/Process/Installer.php
@@ -150,20 +150,15 @@ class Installer
      */
     public function getProcess()
     {
-        switch ($this->type) {
-            case 'github':
-            case 'github-https':
-            case 'bitbucket':
-                if ($this->tree) {
-                    $process = $this->installViaSubtree();
-                }
+        if ($this->type) {
+            if ($this->tree) {
+                $process = $this->installViaSubtree();
+            }
 
-                $process = $this->installViaGit();
-                break;
-
-            default:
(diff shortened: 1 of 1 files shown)


src/Process/Installer.php
```diff
@@ -150,20 +150,15 @@ public function run()
...
-            case 'bitbucket':
-                if ($this->tree) {
-                    $process = $this->installViaSubtree();
-                }
-
-                $process = $this->installViaGit();
-                break;
+        if ($this->type) {
+            if ($this->tree) {
+                $process = $this->installViaSubtree();
```
nWidart: Please early return here which would avoid having to use an else case.

src/Process/Installer.php
```diff
@@ -150,20 +150,15 @@ public function run()
...
+            }
 
-            default:
-                $process = $this->installViaComposer();
-                break;
+            $process = $this->installViaGit();
+        }
+        else {
+            $process = $this->installViaComposer();
```
nWidart: Else no more needed if the previous early return, this can be returned directly too.

src/Process/Installer.php
```diff
@@ -197,11 +192,20 @@ public function getRepoUrl()
...
+
             case 'bitbucket':
                 return "git@bitbucket.org:{$this->name}.git";
 
             default:
-                return null;
+                $parts = explode('.', $this->type);
+                if (end($parts) == "git")
+                    return $this->type;
+                else
```
nWidart: There is no need for an else statement since you early returned. Also please use `{}` around the if.

tomgrv:
Thanks for the remarks.
Please find code cleanup as suggested, and better handling of "type"

nWidart:
Thank you 😄  One last detail, could you put the bracers on the same line as the if conditions, so that the code style is the same for the project.

From this style:
```php
if (filter_var($this->type, FILTER_VALIDATE_URL))
{
    return $this->type;
}		
```


To this, how we do it (PSR-2)
``` php
if (is_null($this->version)) {
    return $this->name . ':dev-master';
}
```

Thanks!

tomgrv:
here we are :-)

nWidart:
Thank you! 👍

tomgrv:
missing call to installViaGit() removed by error during cleanup :-(
now corrected :-)

PR #259, opened 2017-06-29
Revise json cache

tentus:
The project I work on has various instances of the codebase, with modules enabled and disabled according to client needs. To avoid conflicts, we do not have the module.json checked into the project repo, but rather manage those files through some custom logic in Laravel. The changes introduced in 1.25 force modules to have a valid json file when they are initialized, which means that there is never a chance for Laravel to act on any undeployed module configs, since the missing files trigger a File Missing exception.

This PR solves this problem by moving the json requirement back into `json()`, and additionally improves performance by also caching any other files requested.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -26,7 +26,7 @@ class Collection extends BaseCollection
     {
         return array_map(function ($value) {
             if ($value instanceof Module) {
-                return $value->moduleJson->getAttributes();
+                return $value->json()->getAttributes();
             }
 
             return $value instanceof Arrayable ? $value->toArray() : $value;
--- a/src/Module.php
+++ b/src/Module.php
@@ -27,15 +27,16 @@ class Module extends ServiceProvider
     protected $name;
 
     /**
-     * The module path,.
+     * The module path.
      *
      * @var string
      */
     protected $path;
+
     /**
-     * @var Json
+     * @var array of cached Json objects, keyed by filename
      */
-    public $moduleJson;
+    protected $moduleJson = [];
 
     /**
      * The constructor.
@@ -49,7 +50,6 @@ class Module extends ServiceProvider
         parent::__construct($app);
         $this->name = $name;
         $this->path = realpath($path);
-        $this->moduleJson = new Json($this->getPath() . '/module.json', $this->app['files']);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


nWidart:
Hi,

This change was done after multiple blackfire.io profiles.  

Looking at this one https://blackfire.io/profiles/d935c115-104f-47ba-b2b1-ec4610ea7991/graph we can see multiple calls to the `json` method where not always needed.

This changes avoids this logic when reading the module.json file multiple times. 
This profile: https://blackfire.io/profiles/c993acea-0578-4b87-afff-803fc4dc84c9/graph has the changes applied. The call graph is much cleaner in the process without the need for an additional cache.

Not that in the Json class, there's already caching in place.

For this reason I can't revert those changes back, even with more cache.

tentus:
Can you run a profile on my PR? There should be equal or fewer json reads on my branch, I was careful not to undo the benefit the 1.25 release added, just make it so the read of module.json during construct isn't necessary.

nWidart:
It will bring back the irregular calls to the json method since you call it again. Caching there has no use, since the Json class itself has caching already. This was to improve performance without having to rely on caching.

gmox:
It looks like calling `json()` without parameters is effectively the same as accessing the `moduleJson` field. Is the extra call you want to avoid just to simplify branching?

nWidart:
I misread your code, at quick glance I thought you were using laravel caching feature on the json method.

I've ran a profile against your PR, https://blackfire.io/profiles/46d6cfca-ecec-40fb-9828-9c5b13f85da0/graph , it seems pretty decent as well.

nWidart:
Actually, in this profile there are 314 calls to the `json()` method, vs 0 in the current solution.

tentus:
Calling `json()` should be very, very slightly slower than accessing a property directly on most servers, since all it's having to do is set a string and call `Arr::get`, which checks the array is accessible and the value is set. On my local machine, iterating over the two approaches only showed a difference of 0.76 seconds after ten thousand calls.

nWidart:
Very interesting. I think that's an acceptable compromise for added flexibility for your use-case. 😄 

I like this technique you used to have some sort of 'native' caching! 👍

joshbrw:
Just dropping by to say how awesome that `array_get` _cache_ technique is!

PR #272, opened 2017-07-18
Lumen support, fix helpers test

deividaspetraitis:
Support for Lumen:
- Lumen doesn't support boot() methods in providers. In lumens case namespaces and modules is loaded in register() method.
- ....

Fixed helper test.

In future there should be added code boilerplates specific for Lumen instead used Laravel`s. However this pull request is good starting point.


src/LaravelModulesServiceProvider.php
```diff
@@ -43,20 +43,36 @@ public function register()
...
     {
-        $this->app->booted(function ($app) {
-            Stub::setBasePath(__DIR__ . '/Commands/stubs');
+        Stub::setBasePath(__DIR__ . '/Commands/stubs');
 
-            if ($app['modules']->config('stubs.enabled') === true) {
-                Stub::setBasePath($app['modules']->config('stubs.path'));
-            }
-        });
+        switch (get_class($this->app)) {
```
nWidart: Please avoid using a switch statement.

tests/HelpersTest.php
```diff
@@ -30,6 +30,6 @@ public function tearDown()
     /** @test */
     public function it_finds_the_module_path()
     {
-        $this->assertTrue(str_contains(module_path('Blog'), 'modules/Blog'));
+        $this->assertTrue(str_contains(module_path('Blog'), 'Modules/Blog'));
```
nWidart: Why was this needed, tests are green as it was ?

joshbrw:
Any chance that `switch` statement could be refactored out? Potentially something like `$this->isLumen()`?

src/Module.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Nwidart\Modules;
 
+use Illuminate\Container\Container;
 use Illuminate\Foundation\AliasLoader;
 use Illuminate\Foundation\Application;
```
nWidart: If Application isn't used anymore, no need for import.

nWidart:
I think a better way would be to create a service provider dedicated to Lumen only. That way people on Lumen can use that service provider, and the laravel users use the laravel service provider.

deividaspetraitis:
@nWidart good idea. Agree with you about dedicated service provider for Lumen.

src/Module.php
```diff
@@ -170,14 +170,12 @@ public function boot()
...
      * Register module's translation.
-     *
-     * @return void
      */
     protected function registerTranslation()
     {
         $lowerName = $this->getLowerName();
 
-        $langPath = $this->getPath() . "/Resources/lang";
+        $langPath = $this->getPath().'/Resources/lang';
```
nWidart: This change should not have happened.

src/Module.php
```diff
@@ -198,7 +196,7 @@ public function json($file = null)
         }
 
         return array_get($this->moduleJson, $file, function () use ($file) {
-            return $this->moduleJson[$file] = new Json($this->getPath() . '/' . $file, $this->app['files']);
+            return $this->moduleJson[$file] = new Json($this->getPath().'/'.$file, $this->app['files']);
```
nWidart: This change should not have happened.

src/Module.php
```diff
@@ -249,7 +247,7 @@ public function register()
      */
     protected function fireEvent($event)
     {
-        $this->app['events']->fire(sprintf('modules.%s.' . $event, $this->getLowerName()), [$this]);
+        $this->app['events']->fire(sprintf('modules.%s.'.$event, $this->getLowerName()), [$this]);
```
nWidart: This change should not have happened.

src/Module.php
```diff
@@ -279,7 +277,7 @@ protected function registerProviders()
     protected function registerFiles()
     {
         foreach ($this->get('files', []) as $file) {
-            include $this->path . '/' . $file;
+            include $this->path.'/'.$file;
```
nWidart: This change should not have happened.

src/Module.php
```diff
@@ -400,7 +398,7 @@ public function delete()
      */
     public function getExtraPath($path)
     {
-        return $this->getPath() . '/' . $path;
+        return $this->getPath().'/'.$path;
```
nWidart: This change should not have happened.

nWidart:
Please remake a new pull request. You are changing code style that should not have changed.

deividaspetraitis:
That's weird. Not sure why but in other case GrumPHP is failing. 
screenshot_8

nWidart:
Hm I see, that's strange indeed. Will check it out.

nWidart:
aha, for phpcsfixer 2, the config file is .php_cs.dist, and did not update grumphp to use that.

deividaspetraitis:
Rolled back these code style changes and opened new pull request: https://github.com/nWidart/laravel-modules/pull/275

PR #275, opened 2017-07-19
Lumen support

deividaspetraitis:
Support for Lumen:

- Lumen doesn't support boot() methods in providers. Created separate service provider for lumen.
- Requested changes from old pull-request.


nWidart:
The same code style changes are here too. Use the config file of php-cs-fixer to have the correct CS.

deividaspetraitis:
@nWidart should be fine from now.

nWidart:
I don't understand what you're doing.

Is the only difference between the providers where methods are called (register/boot)?

PR #277, opened 2017-07-19
Lumen support

deividaspetraitis:
Support for Lumen:

- In Lumen case namespaces ( haven't tested on laravel since I left it as it is ) must be loaded before registering module provider since it needs some configuration values. Loading these in boot() method will fail to init provider.
- Requested changes from old PR.

### DIFF
--- a/src/LaravelModulesServiceProvider.php
+++ b/src/LaravelModulesServiceProvider.php
@@ -2,37 +2,17 @@
 
 namespace Nwidart\Modules;
 
-use Illuminate\Support\ServiceProvider;
-use Nwidart\Modules\Providers\BootstrapServiceProvider;
-use Nwidart\Modules\Providers\ConsoleServiceProvider;
-use Nwidart\Modules\Providers\ContractsServiceProvider;
 use Nwidart\Modules\Support\Stub;
 
-class LaravelModulesServiceProvider extends ServiceProvider
+class LaravelModulesServiceProvider extends ModulesServiceProvider
 {
-    /**
-     * Indicates if loading of the provider is deferred.
-     *
-     * @var bool
-     */
-    protected $defer = false;
-
     /**
      * Booting the package.
      */
     public function boot()
     {
         $this->registerNamespaces();
-
         $this->registerModules();
-    }
-
-    /**
-     * Register all modules.
-     */
-    protected function registerModules()
-    {
-        $this->app->register(BootstrapServiceProvider::class);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


nWidart:
Thank you for taking the time to resolve the past issues both on your and my end! 😄

deividaspetraitis:
Fantastic! Thanks @nWidart for tips and suggestions!:)

nWidart:
I've changed this https://github.com/nWidart/laravel-modules/commit/c7bad606db27fc67ad44d67f7d066fd8d25e18e0 to use the interfaces, I assume lumen uses these too.

This allows to only have 1 typehint which looks nicer.

deividaspetraitis:
Not really. Lumens application doesn't implement any interface it just extends Container. See: https://github.com/laravel/lumen-framework/blob/5.4/src/Application.php

nWidart:
Hm but that container extends the interface ?

deividaspetraitis:
Unfortunately nope. This interface is available only on Laravel framework and is implemented directly by Laravel`s `Application` class and not by `Container`. Lumen doesn't have such interface: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Foundation/Application.php

nWidart:
Damn ok, so it's not working for you now on lumen ?

Btw did you see lumen isn't maintained anymore? 😛

deividaspetraitis:
Nope. It doesn't, since it requires a class that implements not existing interface.

`Type error: Argument 1 passed to Nwidart\Modules\Repository::__construct() must be an instance of Illuminate\Contracts\Foundation\Application, instance of Laravel\Lumen\Application given, called in /home/vagrant/Code/xxx/public_html/vendor/nwidart/laravel-modules/src/ModulesServiceProvider.php on line 62`

nWidart:
This is annoying as scrutinizer-ci will error on the 2 possible arguments...

deividaspetraitis:
I'm sorry I'm not very familiar with scrutinizer-ci and not sure what to suggest in this case..

deividaspetraitis:
Any updates according this? Still got to use my local version for lumen support.

nWidart:
Sorry I'm not forgetting, still need to change the typehints.

deividaspetraitis:
Could you suggest options how to solve it? I could create a PR for it.

nWidart:
I just had to put back the double type hints you had originally I think since there are no common interfaces between laravel and lumen container / application.

deividaspetraitis:
Yea. I think it's the most reasonable solution for now. Please ping me when you push this change. Also it's important to inject not Laravels `Application` but `Container` instead as it was commited in my PR.

nWidart:
hmm well, it's not clear. In one place Application is used and Containers. Not sure it's worth supporting an abandoned project for all this trouble.

deividaspetraitis:
What do you mean abandoned? Latest version of Lumen is 5.4 ( https://lumen.laravel.com/docs/5.4/releases#5.4.0 ) the same as Laravel. :)

nWidart:
Look at the commits on the lumen repository.

image

deividaspetraitis:
I'm not sure that I understand what you're referring to. Lumen framework has separate repo that is included in composer.json: https://github.com/laravel/lumen-framework/commits/5.4 and is supported as laravel is.

nWidart:
It's the same there though if you look at overall activity. There's also a tweet/post somewhere where Taylor doesn't recommend using Lumen anymore.

deividaspetraitis:
Let's check it: https://github.com/laravel/lumen-framework/issues/636

nWidart:
I see ok, thanks for the heads up 😄 

Could be cool if you can send the PR just to change the docblocks as I'm confused by lumen.

Issue #268, opened 2017-07-16
Hook in the install process


Goopil:
Hello
I'm looking for a simple way to hook in the install / uninstall process.

Is there any events triggered, when command are called ?

This is to have a route registration mechanism to populate menu. Do you have some pointers ?

thanks you in advance.

nWidart:
Hello,

I don't recall having events in the install command (there is no uninstall one). 
This could a good idea to have so I'd accept a PR for this feature.

Goopil:
Ok i will look in to it. Just to be sure, how you see it. It should be a "beforeInstall" "afterInstall" and expose the module and the command (to make prompt if needed) is this implementation ok for you ?

nWidart:
I think we could simply trigger 2 events, in these locations:

- https://github.com/nWidart/laravel-modules/blob/master/src/Process/Installer.php#L133 (before, `ModuleIsInstalling`)
- https://github.com/nWidart/laravel-modules/blob/master/src/Process/Installer.php#L142 (after, `ModuleWasInstalled`)

These events could accept the module name.

Goopil:
Ok i will try somethings, what about enabled & disabled ? Are you also interested ?

nWidart:
Not yet, as that feature will change in the near future.

Goopil:
do you have a roadmap somewhere ?

nWidart:
No roadmap no. Note that PR need start and go to the 1.0 branch, which is for laravel 5.4

Goopil:
You're right, it is a mather of context to have directly the module inside it. So here you'are.

PR #290, opened 2017-07-29
Feature/install events

Goopil:
implements pre & post install events has discussed in #268 

## usage
```php
// App/Providers/AppServiceProvider.php

use Nwidart\Modules\Events\ModuleIsInstalling;
use Nwidart\Modules\Events\ModuleWasInstalled;
...


    public function boot()
    {
        \Event::listen(ModuleIsInstalling::class, function (ModuleIsInstalling $event) {
            $module = $event->getModule();
            $console = $event->getConsole();
        });

        \Event::listen(ModuleWasInstalled::class, function (ModuleWasInstalled $event) {
            $module = $event->getModule();
            $console = $event->getConsole();
        });
    }
```


nWidart:
These events don't need the console, repository, module. Just the module name is enough.

src/Process/Installer.php
```diff
@@ -57,14 +59,14 @@ class Installer
...
      */
     public function __construct($name, $version = null, $type = null, $tree = false)
     {
-        $this->name = $name;
+        $this->name    = $name;
         $this->version = $version;
-        $this->type = $type;
-        $this->tree = $tree;
+        $this->type    = $type;
+        $this->tree    = $tree;
```
nWidart: Code style issue here.

nWidart:
You're also missing blank lines at end of files. https://insight.sensiolabs.com/projects/25320a08-8af4-475e-a23e-3321f55bf8d2/analyses/187

src/Process/Installer.php
Goopil: You are right this is php storm doing it one thing here.

Goopil:
The console should be availlable if you want to prompt a choice. Don't you think ? Ok for the code Style.

src/Events/ModuleBaseEvent.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * The console command instance.
+     *
+     * @var Command
+     */
+    protected $console;
+
+    public function __construct($name, Repository $repository, Command $console)
+    {
+        $this->module = $repository->get($name);
```
nWidart: Please don't use the repository to get the name, when you have the name right here. This is for performance reasons. The repository isn't needed in the event.

Goopil: This is not to have the name but the module itself. Should i do it in the installer instead ?

nWidart: I don't think the event needs the module instance. As I mentioned in the issue, the name is enough.

src/Process/Installer.php
```diff
@@ -134,10 +134,10 @@ public function run()
     {
         $process = $this->getProcess();
         $app = $this->console->getLaravel();
+        $module = $this->repository->get($this->getModuleName());
```
nWidart: Why? The **name** is enough.

Goopil: updated if you still want it

PR #303, opened 2017-08-22
publish language files

chieftaha:
I am suggesting this, to take the benefits of "Overriding Package Language Files" feature, provided by Laravel (As described [here](https://laravel.com/docs/5.4/localization#overriding-package-language-files)).


nWidart:
Hi, 
Thanks but there is no real reason to do this, besides the laravel docs.
Also this is a breaking change.

nWidart:
I'll close this for now, but we'll be able to re-evaluate for `4.0`

PR #307, opened 2017-08-24
adds module name to disable/enable events, further standardizes events

dallincoons:
It would be helpful to be able to hook into a specific event when a module is enabled or disabled. This changes these events to simply call the already existing `fireEvent` which I figured would help standardize the event names.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~5.7",
-    "mockery/mockery": "~0.9",
+    "mockery/mockery": "~0.9.3",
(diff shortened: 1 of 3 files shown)


nWidart:
Hi,

Sorry for the late reply. I like the idea, however, for some reason, the tests are failing on the lower bound dependencies.

dallincoons:
I bumped up Mockery to 0.9.3 because Laravel 5.1 uses Mockery spies in `Illuminate\Foundation\Testing\Concerns\MocksApplicationServices::spy`

nWidart:
Thanks!

PR #322, opened 2017-09-04
Load all the services providers attached to a module... 

mathieutu:
... and write a manifest for each module like Laravel does.

TODO:
- [x] Write some tests about that. Laravel doesn't have any, so need to think about it.
- [ ] Test with Lumen.


nWidart:
Thanks! 👍

PR #328, opened 2017-09-05
Fix wrong path for registerFactories() in prvoider.stub

savander:
Previously, app was search for factories in `Module/Providers/Database/factories` instead of `Module/Database/factories`


nWidart:
Whoops, nice catch. Thanks! 👍

PR #348, opened 2017-09-25
Fix #338, introduce command for creating factories

### LINKED ISSUE #338: Create a module:make-factory [name] [module] command
As of Laravel 5.5 there is a command for making factories.

Kyslik:
So far it does not support `--model` argument; just a stub

Path is not configurable since there are already some places ([service provider stub](https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/scaffold/provider.stub#L99)), where `/Database/factories` path is hard-coded, thus possible BC.

---

Made new issue for content that was here.

### DIFF
--- a/src/Commands/MakeFactoryCommand.php
+++ b/src/Commands/MakeFactoryCommand.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Nwidart\Modules\Commands;
+
+use Illuminate\Support\Str;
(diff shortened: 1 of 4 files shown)


Kyslik:
Hold on with this PR, I will update test same as policy one :)

nWidart:
Thanks!

PR #354, opened 2017-09-26
Renaming command files to use same semantics as Laravel

Kyslik:
This PR:
 - rewords descriptions of commands to use same naming 
    - example: Create a new validation rule...
 - renames files to ThingMakeCommand, now its easier to search for the file (visually) since list will be alphabetically ordered
    - example: MiddlewareMakeCommand.php
 - removes unused parameters sent along to stub class
 - fixes uncovered bug with `getDefaultNamespace()` see 787357c

---

Tests renaming is next, did not want to fiddle with tests while renaming files in src.

### DIFF
--- a/src/Commands/CommandMakeCommand.php
+++ b/src/Commands/CommandMakeCommand.php
@@ -10,7 +10,6 @@ use Symfony\Component\Console\Input\InputOption;
 
 class CommandMakeCommand extends GeneratorCommand
 {
-
     use ModuleCommandTrait;
 
     /**
@@ -34,7 +33,6 @@ class CommandMakeCommand extends GeneratorCommand
      */
     protected $description = 'Generate new Artisan command for the specified module.';
 
-
     /**
      * Get default namespace.
      *
@@ -44,7 +42,6 @@ class CommandMakeCommand extends GeneratorCommand
     {
         return 'Console';
     }
-
 
     /**
      * Get the console command arguments.
@@ -59,7 +56,6 @@ class CommandMakeCommand extends GeneratorCommand
         ];
     }
 
-
     /**
      * Get the console command options.
      *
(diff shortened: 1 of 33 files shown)


nWidart:
Hello,

Thanks for the PR. However, I was about to start refactoring to the `MakeSomethingCommand` structure. 
This makes the files grouped together.

nWidart:
I see you also did a lot of code style changes that aren't needed.

nWidart:
I don't understand the issue about "fixes uncovered bug with getDefaultNamespace() see 787357c"

It never caused an issue in the past, do you have a test making the bug visible?

Kyslik:
I think we should follow how Laravel does it with filenaming (ThingMakeCommand) and maybe use another folder for so-called "generators" but its not needed. Reasoning for this that is that If I search for Middleware (visually, not via search in editor) I can find it easier.

The bug I am talking about in 787357c is that if user sets for lets say for [Notifications](https://github.com/nWidart/laravel-modules/pull/354/commits/787357c362bbf2d0978c5ad14dbe77a8a312fa86#diff-d9c850a8c622c50347c06bb5873d8ab7) in config 'SMS/Notifications', the `getDefaultPath()` is going to generate wrong path. In other words I did not look at the detail I just concluded that half of commands had default namespace hard-coded and half did not.

nWidart:
Yes, I looked at what laravel does, but I don't think we need to copy laravel in regards to class names.

Kyslik:
And code style changes is from PHPstorm cmd+alt+L set to PSR2. I just "styled" those that I touched.

I am following Laravel's conventions because this package is targeted for Laravel users not for some other framework users.

nWidart:
The code style is not the one from phpstorm, but the one from phpcsfixer.

There's no need to follow laravel conventions.

Kyslik:
Changed styling using php-cs-fixer; commit message should be "Using .php_cs.dist".

nWidart:
Thank you 👍

nWidart:
Please hold of the tests renaming, I'll handle that. 😄

Kyslik:
Sure thing :D

And thanks too :).

PR #355, opened 2017-09-26
Lumen 5.5.x support ( removed mixed Laravel and Lumen dependencies )

deividaspetraitis:
Since laravels dependency is required in composer.json it lead to following error: https://github.com/laravel/lumen-framework/issues/655 that is caused by mixed dependencies.

Because package supports both: Lumen and Laravel I think it's good idea to move laravel requirement to dev section only for testing purposes.


nWidart:
Hello, thanks I was thinking about doing this as well.

Quick comment, it seems like you forgot to import the alias loader in the laravel/modules class.

nWidart:
There has been a pretty big change on the repository, specifically class names that have changed. 

Could you re-submit your PR with the up to date upstream and the fix to the class import of the AliasLoader please?

Thanks!

deividaspetraitis:
Sure, let me check. Just strange that all tests passed.

nWidart:
Indeed, but the error was shown on sensiolabs insights

PR #357, opened 2017-09-26
Lumen 5.5.x support ( Removed mixed Laravel and Lumen dependencies )

deividaspetraitis:
See https://github.com/nWidart/laravel-modules/pull/355

-  PR with the up to date upstream
-  Fixed class import of the AliasLoader

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,14 +18,14 @@
     }
   ],
   "require": {
-    "php": ">=7.0",
-    "laravel/framework": "5.5.*"
+    "php": ">=7.0"
   },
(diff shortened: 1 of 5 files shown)


nWidart:
Thanks!

What's this public_path helper you added? Is it the same as laravel's one, for lumen?

deividaspetraitis:
That's right. By default lumen doen't have such helper function and throws errors because of this that's why I added it.

nWidart:
I see, thanks.

deividaspetraitis:
Thanks!

Issue #366, opened 2017-10-03
custom default generate file and folder in module


sonminh18:
how to custom default generate file and folder in module? 
because i think some file and folder not necessary,so i want to remove it when i make a new module. 
I have tried to change config in vendor/nwidart/laravel-module/config/config.php, and published by command "php artisan vendor:publish --provider="Nwidart\Modules\LaravelModulesServiceProvider""
but nothing happens
Thanks!

IllyaMoskvin:
As a general rule, you should avoid changing things in the `vendor` folder, since your changes will be overwritten whenever you run `composer update`. That said, I just got this working today, so it's pretty fresh in my mind, and I'll try to share what worked for me.

I found the documentation to be slightly misleading at the beginning. It claims that...

> The package will automatically register a service provider and alias.

...but I found that I had to do this manually.

First, go to `config/app.php`. Find the `providers` and `aliases` settings towards the bottom of the file, and add the following lines to them, respectively:

```
'providers' => [

    // Add this after all of your other providers...
    Nwidart\Modules\LaravelModulesServiceProvider::class,

],

'aliases' => [

    // Add this after all of your other aliases...
    'Module' => Nwidart\Modules\Facades\Module::class,

],
```

Now, you should be able to run the command you tried before:

```
php artisan vendor:publish --provider="Nwidart\Modules\LaravelModulesServiceProvider"
```

This will create the file `config/modules.php`. Its primarily purpose is to tweak module generation. Read the comments in that doc, they should be helpful for answering your question.

In particular, take a look at `stubs.files`. You can comment out any of the lines in that array, and those files will no longer be generated. For example:

```
    'stubs' => [
        'files' => [
            // Start simply registers the routes...
            'start' => 'start.php',
            'routes' => 'Http/routes.php',
            // I don't want views in my modules, so I disabled them by default
            // 'views/index' => 'Resources/views/index.blade.php',
            // 'views/master' => 'Resources/views/layouts/master.blade.php',
            // Unfortunately, config currently gets generated even if this is disabled
            'scaffold/config' => 'Config/config.php',
            'composer' => 'composer.json',
        ],
    ]
```

Additionally, scroll down a bit to the `paths.generator` section. You should see a list of directories here. If you set any of them to `false` (with a couple exceptions), those directories will not be generated. Here is my config, for comparison:

```
    'generator' => [
            'assets' => false, // 'Assets',
            // Unfortunately config gets generated even if this is disabled
            'config' => 'Config',
[...]

PR #367, opened 2017-10-03
Clarify installation instructions for Laravel 5.4

IllyaMoskvin:
I haven't yet gotten around to updating my project to Laravel 5.5, and I haven't been following Laravel news, so I wasn't aware of these developments:

https://laravel-news.com/package-auto-discovery
https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518

This PR essentially reverses 7a8b50b, while differentiating between Laravel 5.5 and 5.4. I think it's worthwhile to include, since AFACT, it's the only tidbit preventing its compatibility with version `2.0^` of this package. Plus, I think it's still relevant to whomever might be using version `1.0^`. IIRC, during my installation today, I found the solution by stumbling over the legacy pingpong labs repo. Might as well keep it here until Laravel 5.4 is at least one more version out-of-date.

I also added a teaser re: config options to the readme, both due to today's inquiry (#366) and because I first heard about this project via an (outdated) discussion, where one party was critiquing the package for generating too many folders by default. I think people will want to know that functionality is easily tweakable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,25 @@ To install through Composer, by run the following command:
 composer require nwidart/laravel-modules
 ```
 
-The package will automatically register a service provider and alias.
+In Laravel 5.5+, this package will automatically register a service provider and alias. If you are running Laravel 5.4, you will need to add them manually by editing your `config/app.php` as follows:
+
+``` php
+'providers' => [
+  Nwidart\Modules\LaravelModulesServiceProvider::class,
+],
+
+'aliases' => [
+  'Module' => Nwidart\Modules\Facades\Module::class,
+],
+```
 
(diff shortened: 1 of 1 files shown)


nWidart:
Thanks, but this isn't needed. The instructions for <5.5 are on the 1.0 branch.

IllyaMoskvin:
Good call, I did not notice that existed. Would you be down to make the "^1.0" in the table into a link to [tree/1.0](https://github.com/nWidart/laravel-modules/tree/1.0)? Could help newbies like me.

nWidart:
Hm I guess that could be a link yes. Though no need for the `tree/` part.

PR #384, opened 2017-10-09
Added migration status

### LINKED ISSUE #128: Migration Status of modules doesn't show up in migrate:status
Hey folks,

while I am developing each module I'd like to see the migration status of each module seperatly. But they doesn't even show up in the default migration:status command without publishing it.
This makes handling the migration hard if you need to rollback and migrate again and have a couple modules to have all in mind. 
 
 
Is this the way it should work or is there an easy way to show all of the migration in one place?
Sofar the migration Table in the DB have all records, but only the avaible migration files in the default /database/migrations folder are shown up.

sazo:
Fixes #128

### DIFF
--- a/src/Commands/MigrateStatusCommand.php
+++ b/src/Commands/MigrateStatusCommand.php
@@ -90,5 +90,4 @@ class MigrateStatusCommand extends Command
             array('database', null, InputOption::VALUE_OPTIONAL, 'The database connection to use.'),
         );
     }
-
 }
--- a/src/Commands/MigrateStatusCommand.php
+++ b/src/Commands/MigrateStatusCommand.php
@@ -34,7 +34,7 @@ class MigrateStatusCommand extends Command
      *
      * @return mixed
      */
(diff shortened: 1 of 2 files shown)


src/Commands/MigrateStatusCommand.php
```diff
@@ -0,0 +1,93 @@
...
+     * @var \Nwidart\Modules\Repository
+     */
+    protected $module;
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function fire()
```
nWidart: This has to be `handle()`

nWidart:
Hi and thanks for the contribution. Looks good to me besides renaming fire to handle.
Also please use short array syntax.

sazo:
@nWidart Your welcome! :) Love the project - makes big code bases alot more maintainable. 

PS. have correct according to your comments.

nWidart:
Awesome, thanks!

PR #391, opened 2017-10-13
Improved support for Lumen

ReinierH:
Added extra helper function and updated the readme

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -9,6 +9,20 @@ if (! function_exists('module_path')) {
     }
 }
 
+if ( ! function_exists('config_path'))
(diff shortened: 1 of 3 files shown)


src/helpers.php
```diff
@@ -9,6 +9,20 @@ function module_path($name)
     }
 }
 
+if ( ! function_exists('config_path'))
+{
```
nWidart: Please run phpcsfixer. The brace should be on the same line.

ReinierH: done

nWidart:
Thanks. 👍

PR #411, opened 2017-11-03
Make gitkeep optional

balping:
I'm aware what gitkeep files are for, yet I'd like to disable them. This PR adds an option to turn their generation off.


nWidart:
Hello & thanks for your contribution.

I agree they can be annoying at times.
I think however a better solution would be to not generate them in a non-empty folder.

This is a good first step. 👍

Issue #400, opened 2017-10-24
LoadTranslationsFrom


jnbn:
Instead of having registerTranslations method on provider stub; I think it might be much better if we use

```
$this->loadTranslationsFrom(__DIR__ .'/../$PATH_LANG$', '$LOWER_NAME$')
```

as Laravel supports overriding package language files. https://laravel.com/docs/5.5/localization#overriding-package-language-files

What do you think about it?

nWidart:
This is already the case https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/scaffold/provider.stub#L81

jnbn:
I mean instead of having
```
$langPath = resource_path('lang/modules/$LOWER_NAME$');

        if (is_dir($langPath)) {
            $this->loadTranslationsFrom($langPath, '$LOWER_NAME$');
        } else {
            $this->loadTranslationsFrom(__DIR__ .'/../$PATH_LANG$', '$LOWER_NAME$');
        }
```
just `$this->loadTranslationsFrom($langPath, '$LOWER_NAME$');` seems enough for anyone as Laravel automatically looks for lang/vendor/modulename as it's said on documentation.

`Some packages may ship with their own language files. Instead of changing the package's core files to tweak these lines, you may override them by placing files in the  resources/lang/vendor/{package}/{locale} directory.

So, for example, if you need to override the English translation strings in messages.php for a package named skyrim/hearthfire, you should place a language file at:  resources/lang/vendor/hearthfire/en/messages.php. Within this file, you should only define the translation strings you wish to override. Any translation strings you don't override will still be loaded from the package's original language files.`

jnbn:
Briefly we're referring 'lang/modules/modulename' unnecessarily because it's already set as a fallback to lang/vendor/modulename by Laravel with the loadTranslationsFrom method.

nWidart:
Oh I see what you mean, will check it out, thanks!

enextiarg:
Hi Nico,

Maybe helps, i added in service provider the following lines in order to add json language package from module. 

if (is_dir($langPath)) {
            $this->loadJsonTranslationsFrom($langPath, 'lang');
        } else {
            $this->loadJsonTranslationsFrom(__DIR__ .'/../Resources/lang', 'lang');
        }

PR #412, opened 2017-11-04
loading JSON translations

Eric-all:


Eric-all:
@nWidart lock https://github.com/nWidart/laravel-modules/issues/400

Eric-all:
@nWidart  look , https://github.com/laravel/framework/pull/20599   😃

nWidart:
Thanks, though this seems to be adding I/O operations. Don't you think it's easy for someone who wants json translations to add it?

Maybe it could be documented on the docs.

src/Module.php
```diff
@@ -190,6 +190,7 @@ protected function registerTranslation()
 
         if (is_dir($langPath)) {
             $this->loadTranslationsFrom($langPath, $lowerName);
+            $this->loadJSONTranslationsFrom($langPath);
```
mikemand: This should be `loadJsonTranslationsFrom`. JSON is not capitalized: [Illuminate/Support/ServiceProvider.php#L107-L116](https://github.com/laravel/framework/blob/dfcb8e372e7efb0e95c5844571c0701e91b7f2e6/src/Illuminate/Support/ServiceProvider.php#L107-L116)

nWidart:
Thank you for the contribution, but this is something that's easily added manually and avoids extra load.

Issue #424, opened 2017-11-17
RouteServiceProvider route bind


angujo:
Not sure whether this has been asked before or not, but bear with me.
Have generated a routeservice provider for my module.

Have setup Route bind on the before function as indicated below;
```
public function before(Router $router)
    {
        \Route::bind('worker', function ($hid) {
            return Worker::find(($hid)) or JsonResult::exception('Invalid Worker');
        });
    }
```

The above doesn't seem to be reached, even adding a breakpoint in teh function doesn't get called.
The service provider was generated via `php artisan module:route-provider Abcm`
What am I missing?

Thanks.

angujo:
Hi Good People,

Is there anyone who has been able to work with the RouteServiceProvider in a module?
Literally stuck at the moment. Have tried registering it on  the `ModuleServiceProvider` `register` function, but all in vain.

How do I register it or any pointer to documentation on the same?

Regards

nWidart:
Hi,

Yes I'm using the RouteSP one very module I use. Look at the modules here: https://github.com/idavoll

mfullbrook:
Hi @nWidart, superb library just started to use it.  I've checked Setting and Core from https://github.com/idavoll and you are not using the `before()` hook.  I'm grepping laravel-modules and laravel for something that calls `before()` and cannot find anything.

The hook i usually use in service providers is `boot()`.  Should before be boot?  Happy to submit a PR if you agree.

nWidart:
Thank you :)

Hm I haven't heard about that before hook, I've always just used `register` and `boot`.

mfullbrook:
https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/route-provider.stub#L25

nWidart:
Ah I see you mean on the routeSP. Yea you're free to use anything you want in there.

angujo:
@mfullbrook Thanks for pinpointing the issue. It has baffled me for hours now.
Sure, when I changed it from `before(Router $router)` to `boot()` it worked.

@nWidart Thanks for a great library.

nWidart:
You shouldn't have to do that though. The sounds like you haven't extended the laravel route service provider which has the before method.

angujo:
I don't think there's a before() function in RouteServiceProvider:

https://laravel.com/api/5.5/Illuminate/Foundation/Support/Providers/RouteServiceProvider.html

Check the API link.

mfullbrook:
[...]

PR #425, opened 2017-11-17
RouteProvider stub - use boot() instead of before()

mfullbrook:
Modules\XXX\Providers\RouteServiceProvider is stubbed with `before()` that isn't actually called. Changing this to `boot()`.

Discussed in https://github.com/nWidart/laravel-modules/issues/424


nWidart:
Thank you! 👍

PR #430, opened 2017-11-19
Route Provider & Config files.

wffranco:
* New route style like laravel 5.3+ (folder with api & web routes separated).

* Fixed routes:
   * removed from the master service provider
   * added a route service provider

* Allow multiple config files in a module. All will be append to the master config array.


config/config.php
```diff
@@ -92,19 +94,19 @@
         | Set the generate key to false to not generate that folder
         */
         'generator' => [
```
mikemand: Why did you change the case on these?

nWidart:
Hello,

Thank you for the contribution and the time you took to make this pull request.
However, I can't accept this as it's unclear with it provides, while also being a breaking change.

PR #431, opened 2017-11-19
Cache and vendor directory

### LINKED ISSUE #429: Cache and vendor dir
Hello,

I don't really understand the purpose of this code segment:

https://github.com/nWidart/laravel-modules/blob/8e86c2b0f0fa37b4684883c996d01c15bd9862e8/src/Laravel/Repository.php#L37-L48

If cache is enabled, it loops through all modules and replaces the path of each module. Why? The new path might or might not be the real one.

For instance I have a module in the vendor folder that is loaded correctly if caching is turned off.

Running `module:list` correctly finds it:

```
+------+---------+-------+------------------------------------------+
| Name | Status  | Order | Path                                     |
+------+---------+-------+------------------------------------------+
| Auth | Enabled | 0     | /project-root/vendor/balping/auth-module |
+------+---------+-------+------------------------------------------+
```

But when I enable caching, the above code tries to cache the path as `project-root/Modules/Auth` which is obviously wrong!

This leads to this error later:

```
[Illuminate\Contracts\Filesystem\FileNotFoundException]  
  File does not exist at path /module.json  
```

So my question is: what's the purpose of overriding the path?  @nWidart said several times in different issues that people use modules from the vendor dir. But do they do that with caching enabled? I haven't found any useful example on that.

balping:
This seems to fix #429

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -26,7 +26,9 @@ class Collection extends BaseCollection
     {
         return array_map(function ($value) {
             if ($value instanceof Module) {
-                return $value->json()->getAttributes();
+                $attributes = $value->json()->getAttributes();
+                $attributes["path"] = $value->getPath();
+                return $attributes;
             }
 
             return $value instanceof Arrayable ? $value->toArray() : $value;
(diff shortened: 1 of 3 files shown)


nWidart:
Thank you! 👍

PR #432, opened 2017-11-19
Added .editorconfig file

joaorobertopb:
I suggest using .editorconfig in this package.


nWidart:
Hello,

Thank you for the idea, but we're already using [phpcsfixer](https://github.com/nWidart/laravel-modules/blob/master/.php_cs.dist) for this purpose.

PR #433, opened 2017-11-19
Added .gitattributes file

joaorobertopb:
This way, the files would be available in the repository but it would not be downloaded when the package is required by another project.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,5 @@
 /.php_cs.dist       export-ignore
 /.scrutinizer.yml   export-ignore
 /.travis.yml        export-ignore
-/CHANGELOG.md       export-ignore
 /phpunit.xml.dist   export-ignore
 /tests              export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,13 @@
...
+# https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html
+
+# This way, the files would be available in the repository but it would not be downloaded when the package is required by another project.
+
+/.gitattributes     export-ignore
+/.gitignore         export-ignore
+/.php_cs.dist       export-ignore
+/.scrutinizer.yml   export-ignore
+/.travis.yml        export-ignore
+/CHANGELOG.md       export-ignore
```
Meroje: Please don't remove documentation

nWidart:
Looks good besides the point @Meroje made. Please don't ignore the changelog file.

nWidart:
Thanks!

PR #438, opened 2017-11-25
Remove code duplication

balping:
`Lumen/Repository.php` and `Laravel/Repository.php` were exact copies. It was unnecessary code duplication. I  moved `scan` and `formatChached` function back to the abstract parent repository.


nWidart:
Sorry for late reply. Thank you for the contribution! 👍

Issue #423, opened 2017-11-13
Module does not exist!


rafapaulino:
I published a module in the packagist, however at the time of installing the module it does not work and generates this error:

  [Nwidart\Modules\Exceptions\ModuleNotFoundException]
  Module [Blog] does not exist!

How do I solve this? I tried with the command: php artisan module:install rafapaulino/blog

nWidart:
Hello,

Did you try to require it with composer using `composer require`? You may also need to specify to load it in the configuration and set the vendor config to true in the config file.

Issue #417, opened 2017-11-08
Cannot define custom scan path not ending with '/*'


aleron75:
Hello,
since `Nwidart\Modules\Repository\getScanPaths()` always adds a trailing `/*` to each path (not already ending with `/*`), I am not able to register a custom scan path like, for example:

    'scan' => [
        'enabled' => true,
        'paths' => [
            base_path('vendor/*/*/src'),
        ],
    ],

Thus, for example in Laravel, the `Nwidart\Modules\Laravel\Repository::scan()` method will fail at finding the `module.json` under my path, because the `$path` used here:

`$manifests = $this->app['files']->glob("{$path}/module.json");`

will be `/var/www/html/vendor/*/*/src/*/module.json` instead of the correct one, which is `/var/www/html/vendor/*/*/src/module.json`.

Since it doesn't seem that the `Nwidart\Modules\Repository\getScanPaths()` is used elsewhere, why the `/*` is always added at the end of the path?

Workaround: I use `'vendor/*/*/'` as base path but this brings to scanning more directories than needed and could find other `module.json` files.

Thank you

nWidart:
Hello,

Interesting use-case.
This was recently added since version 2.6 to allow loading modules in subfolders.
Ideally this path needs to point to the parent folder, where the modules (plural) can live.

jnbn:
Maybe instead of using str_finish we may consider using a regexp or something to allow multiple wildcard subdirectories?

nWidart:
This was just tagged in 2.7.
Sorry for the delay!

PR #441, opened 2017-11-29
fixed modules don't load from vendor folder

yemenifree:
by default, scan path includes all modules from vendor folder via `vendor/*/*` path

```php
    'scan' => [
        'enabled' => false,
        'paths' => [
            base_path('vendor/*/*'),
        ],
    ],
```

But `str_finish` convert the path to `vendor/*` because it removes all `/*` from path to force add only one `/*` in the end, and this cause module doesn’t load from vendor folder, so this PR to check if path end with `/*` already skip adding `/*` to the end.

### DIFF
--- a/src/Repository.php
+++ b/src/Repository.php
@@ -105,8 +105,9 @@ abstract class Repository implements RepositoryInterface, Countable
         if ($this->config('scan.enabled')) {
             $paths = array_merge($paths, $this->config('scan.paths'));
         }
+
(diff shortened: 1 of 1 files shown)


yemenifree:
This fixed [#423](https://github.com/nWidart/laravel-modules/issues/423) [#417](https://github.com/nWidart/laravel-modules/issues/417)

nWidart:
Thank you very much!

👍

PR #444, opened 2017-12-02
include files while booting app.

yemenifree:
In some situations, files of the module need to access to config of the module, but right now files include in the register method when config of modules doesn’t load yet ( it's will be available only if Laravel config is cached)

So, I think files are best to include on boot function to ensure config of modules will be available.

any suggestion is welcome.

### DIFF
--- a/src/Module.php
+++ b/src/Module.php
@@ -174,6 +174,8 @@ abstract class Module extends ServiceProvider
             $this->registerTranslation();
         }
 
+        $this->registerFiles();
+
         $this->fireEvent('boot');
     }
 
@@ -245,8 +247,6 @@ abstract class Module extends ServiceProvider
         $this->registerAliases();
 
         $this->registerProviders();
-
-        $this->registerFiles();
 
         $this->fireEvent('register');
     }


nWidart:
Hello & thank you for the contribution.

This was suggested before but is not compatible with [asgardcms](https://github.com/asgardcms/platform)

yemenifree:
@nWidart what if we add an option to modules config to load files on register or boot, and by default load by register method?

nWidart:
Hm, that would be doable indeed.

yemenifree:
@nWidart check changes now, I can also make this auto compatible by check if modules run on asgardcms or not and force load from a register for asgardcms.

yemenifree:
commit [d717845a59f473d4d6030acc71d661fbd03ce815](https://github.com/yemenifree/laravel-modules/commit/d717845a59f473d4d6030acc71d661fbd03ce815)

nWidart:
That looks good 👍

yemenifree:
@nWidart see #445

PR #451, opened 2017-12-19
use default stub if override not exists

yemenifree:
If somebody overrides stub path in the config, he should copy new stubs with every new release of the package to add missing stub,

This PR fixes this problem by check if stub file exists on override path use it or use default stub form package folder as a fallback.

I add some test unit for that to make sure everything is work.

thanks


nWidart:
Good idea! Thanks for the contribution. 👍

PR #459, opened 2018-01-13
Fix non well formed numeric value

filtration:
When a row in the database doesn't exist, you get a 

      A non well formed numeric value encountered

This is because you're catching an exception rather than throwable


nWidart:
Thank you! 👍

PR #483, opened 2018-02-16
Fix module:publish-config command

FaZeRs:


nWidart:
Oh nice thank you! Almost forgot about those.

PR #485, opened 2018-02-17
Route serviceprovider

rjvandoesburg:
Hello~

In Laravel 5.3 there was a change in how routes were loaded (see: https://laravel.com/docs/5.3/upgrade#upgrade-5.3.0)

However when using this package I'd always find myself re-writing the routes in the modules to reflect the same changes as in the framework.

Thus I decided to create a PR reflecting the changes from the framework.
Old modules will still work as they still load `start.php` from the `module.json`.

When running `php artisan module:make {name}` these are the new changes:
- A `Providers/RouteServiceProvider.php` is now created and referenced in `Providers/{name}ServiceProvider.php`
- `Http/routes.php` has been moved to `Routes/web.php` and is now loaded from `Providers/RouteServiceProvider.php`
- `start.php` is no longer created and referenced in `module.json` 
- Route cache check in `Routes/web.php` has been removed as this is handled by the RouteServiceProvider

Updated the tests to reflect the new changes. Now all tests success on my and but because I don't know the ins and outs of the module I might have forgotten something. If so, please let me know so I can fix this.

Now this is all on the basis this is a change you are willing to add to the package of course :)


nWidart:
Hello,

Thanks for the contribution! Those route changes are optional, however. Where routes are defined is not imposed by the framework.

On top of this, laravel 5.3 is not supported anymore.

In the future try create an issue first in order to discuss changes, and not wasting time on your end.

rjvandoesburg:
No it is not imposed by the framework, however I merely meant that in 5.3 there was a change to the routes and this PR applies those same changes to the modules.

rjvandoesburg:
@bloggervista I never stated anything was wrong with the code as it is now. I just felt an update on how routes are defined could be updated. Laravel didn't make the change themselves for nothing.

My opinion was if someone starts a new Laravel project from scratch and then use this module, the way routes are defined would be the same as Laravel allowing them to follow the Laravel docs to continue onwards.

Seeing @nWidart you haven't rejected the PR yet, are you considering this change or should I create and Issue asking people if this is a change they'd like to see?

If you don't plan on accepting the PR, just decline it and don't leave me hanging here.

ps. It would seem referring back to L5.3 only caused more confusion than support for this change :/

bloggervista:
yes i know what you mean.  

I also think better to place routes inside routes folder instead of http because i think routes is the first place everybody goes . I guess that was the reason laravel moved routes from app/http/routes to routes/web .

I guess this pr is good :)

nWidart:
Hi,

Sorry, it took so long to reply, I wasn't sure if I wanted this change or not.
After reflection, it does look nice.

Could you please fix the conflicts, I think it's easy to do, someone removed the start file before.
I not I'll take care of it.

Thanks.

rjvandoesburg:
Cheers! I'll have a look at them next week :)

rjvandoesburg:
Sorry it took a bit longer but I've fixed the merge conflicts!

nWidart:
Thanks! Since this is a breaking change, I've created the `3.0`, and this will be on master, which will become 4.0 with laravel 5.7 support.

nWidart:
Master branch on laravel 5.7 is failing 😢

PR #497, opened 2018-02-28
Fix #496 and include the function return.

### LINKED ISSUE #496: References for nothing
The line quoted below appears in several files and references to _something_ that **no longer exists**.

https://github.com/nWidart/laravel-modules/blob/a68dd230d3ba6fd95cb55e4da95a6fd2bdc4e7e3/src/Commands/stubs/scaffold/provider.stub#L94

tkovs:


nWidart:
Thank you!

PR #500, opened 2018-03-04
Allow symlinks in module path

balping:
I couldn't figure out what was gravitano's rational behind using `realpath()` when constructing a `Module` class. It seems unnecessary to me. It wouldn't be an issue though, but this caused me trouble when I added modules behind symlinks, as `realpath` returns an absolute path with resolved symlinks. Removing `realpath` doesn't seem to break anything, but solves my issue.


nWidart:
Sorry for the late reply! Thank you!

PR #505, opened 2018-03-12
Made start.stub more Laravel-like.

BrekiTomasson:
Flag comments always have three lines, with every line being three characters shorter than the one before it.


nWidart:
Thank you!

PR #510, opened 2018-03-21
Laravel Mix

KABBOUCHI:
Edit:
 
Add `webpack.mix.js` and `package.json` stub files

```bash
php artisan module:make Blog && cd Modules/Blog && npm install
```

Every module has its own dependencies, when you run `npm run dev` or `npm run prod` it will copy the files to the route project `/public/modules` after finish compiling

```html
<link rel="stylesheet" href="{{ mix('css/app.css','modules/blog') }}">
```
```html
<script src="{{ mix('js/app.js', 'modules/blog') }}"></script>
```


mikemand:
> Register single module
How does this work? Does it run a module's Mix automatically? Does it use `node_modules` from that module, or does it use them from the root of your project? How would you install dependencies for modules?

I'm currently using Gulp to run Mix for my modules, and it's a huge ugly mess. If this removes that step, I would give this PR a 💯.

KABBOUCHI:
>How does this work? Does it run a module's Mix automatically?

Yes

>Does it use node_modules from that module, or does it use them from the root of your project

From the root project.  
 

There a another way, but you need to run ``npm run dev/watch`` from the module directory.
Compile and output the result in the module dir then copy it to the public path.

image

```html
<link rel="stylesheet" href="{{ mix('css/app.css','modules/blog') }}">
```
```html
<script src="{{ mix('js/app.js', 'modules/blog') }}"></script>
```

I got this idea from [Horizon](https://github.com/laravel/horizon/blob/1.0/webpack.mix.js)

I think the second approach its better and doesn't require from the user to add extra config manually

nWidart:
Thank you for your work on this. It looks very good!

It would be nice if you could prepare some docs too. Either on the [documentation site](https://github.com/nWidart/nwidart-docs) directly or as a comment on this PR.

👍

bloggervista:
Yeah i guess mix should be used as gulp is too messy :(

nWidart:
Thanks again!

balping:
Great PR!

In case you'd like to integrate module assets with other modules' assets and/or run mix from the root project and/or would like to avoid running npm for each module, I recommend this package: https://github.com/balping/asset-bundler

KABBOUCHI:
@mikemand 
```js
if(document.getElementById('app') && ! document.getElementById('app').__vue__)
{
   new Vue({ 'el' : '#app', ... });
}
```

bloggervista:
@balping  thanks for that package. Great work man!!!

mikemand:
@KABBOUCHI I already do check `if(document.getElementById('...'))`, but the problem is I have a "main" Vue instance (that AsgardCms provides), that mounts to `#app`, and then some of my modules have their own Vue instances that try to mount to their own elements that happen to be inside of `#app`. If I move everything to the main `app.js` file, then my Vuex store fails to initialize for some reason.

madsem:
wow man! This is awesome @nWidart @KABBOUCHI right when I need it most I find Laravel Modules and this PR in particular is exactly what I needed. Awesome work guys, thanks a lot. Will try my best to contribute to this project.

KABBOUCHI:
@mikemand this my setup for multiple `vue` instances same on the same element

`primary.js` => Main Dashboard ( html/css/js)
```js
window.Vue = require('vue');
window.events = new Vue({name: 'events'});

const app = new Vue({
    el: '#app'
});

events.$on('loaded', function () {
   app.$forceUpdate();
})

```

`secondary.js`
```js
if (!window.Vue) {
    window.Vue = require('vue');
}

if (!window.events) {
    window.events = new Vue({name: 'events'});
}

Vue.component('component','....');

events.$emit('loaded', {
    module: 'school'
});

// if `primary.js` dosen't exists.
if (document.getElementById('app') && !document.getElementById('app').__vue__) {
    const app = new Vue({
        el: '#app'
    });
}
```

madsem:
@KABBOUCHI would it be a good idea to integrate themes in this as well?

Instead of loading files statically like this
```php
    'assets/js/app' => 'Resources/assets/js/app.js',
    'assets/sass/app' => 'Resources/assets/sass/app.scss',
```

Maybe we could load all .js, .scss & .css files under ```/Resources/assets/*``` recursively also in subfolders, and then copy everything to ```/public/```?

That would enable us to have themes for each module as well

KABBOUCHI:
@madsem  please check this #521, this PR has some issues.

you can't copy everything from ``assets``, you should copy the compiled version

if you need only to write vanilla css/js you can use [`mix.copy()`](https://github.com/JeffreyWay/laravel-mix/blob/master/setup/webpack.mix.js)

you can add/change entries in `webpack.mix.js`

```js
mix.sass( __dirname + '/Resources/assets/sass/themes/default/app.scss', 'css/blog-default.css');
```

madsem:
@KABBOUCHI that's not what I meant.

I was asking if it might be a good idea to have this PR do what I wrote automatically.

Process the entire folder structure in /assets/ recursively in the webpack.mix.js and then copy the entire folder incl. child folders to /public :)

PR #531, opened 2018-05-15
Added an option to prefix table migrations with the module's name.

lexxyungcarter:
## Current
Command
```php
// create a Post model with its migration for the Blog module
php artisan module:make-model -m Post Blog
```
Output
```php
// Post model
<?php

namespace Modules\Post\Entities;

use Illuminate\Database\Eloquent\Model;

class Post extends Model
{
    protected $fillable = [];
}


// migration  2018_05_15_204011_create_posts_table
<?php

use Illuminate\Support\Facades\Schema;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Migrations\Migration;

class CreatePostsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('posts', function (Blueprint $table) {
            $table->increments('id');

            $table->timestamps();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('posts');
    }
}

```

## PR Change
Command
```php
// create a Post model with its migration for the Blog module, but prefix the module's name
php artisan module:make-model Post -pm Post Blog
```
Output
```php
// Post model
<?php

namespace Modules\Blog\Entities;

use Illuminate\Database\Eloquent\Model;

class Post extends Model
{
    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [];

    /**
     * The table associated with the model.
     *
     * @var string
     */
    protected $table = "blog_posts";

    // 
}


// migration file 2018_05_15_204011_create_blog_posts_table
<?php

use Illuminate\Support\Facades\Schema;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Migrations\Migration;

class CreateBlogPostsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('blog_posts', function (Blueprint $table) {
            $table->increments('id');

            $table->timestamps();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('blog_posts');
    }
}

```

### DIFF
--- a/src/Commands/ModelMakeCommand.php
+++ b/src/Commands/ModelMakeCommand.php
@@ -86,6 +86,7 @@ class ModelMakeCommand extends GeneratorCommand
         return [
             ['fillable', null, InputOption::VALUE_OPTIONAL, 'The fillable attributes.', null],
             ['migration', 'm', InputOption::VALUE_NONE, 'Flag to create associated migrations', null],
+            ['prefix', 'p', InputOption::VALUE_NONE, 'Flag to create associated migrations and prefix with module name', null],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


nWidart:
Hello & thank you for the contribution.

I'm not sure what value this brings. Yes, you can prefix tables but that's easily done on edit too, without needing more code.

Don't hesitate to discuss things via an issue first to avoid having unnecessary work 😄 .

twf-nikhila:
@lexxyungcarter @nWidart this might not be added to the module, but having a module name as a prefix is always a good practice. Helps to know the tables which belongs to a particular module.

nWidart:
AsgardCMS also prefixes the migrations with the module name. No need to fork the package, you can "wrap" the code generation.

PR #534, opened 2018-05-18
fix artisan command module:migrate-refresh

huydang284:
Hi @nWidart,

There is a small issue in command "module:migrate-refresh". I fixed it.
Please merge.

Regrads,
Huy Dang

### DIFF
--- a/src/Commands/MigrateRefreshCommand.php
+++ b/src/Commands/MigrateRefreshCommand.php
@@ -81,7 +81,7 @@ class MigrateRefreshCommand extends Command
 
         $module = app('modules')->find($module);
 
-        if ($module === null) {
+        if ($module === null || is_string($module)) {
             return $module;
         }
 


nWidart:
Hello,

Could you explain when this wouldn't be a string?

Thanks!

huydang284:
Your code in `\Nwidart\Modules\Commands\MigrateRefreshCommand::getModuleName`:
```
public function getModuleName()
{
        $module = $this->argument('module');

        app('modules')->find($module);

        if ($module === null) {
            return $module;
        }

        return $module->getStudlyName();
}
```

Above code is missing assignment `$module = app('modules')->find($module);` so `$module` is string, error will be shown when you call function `getStudlyName` from a string.

So below code will be right (not my pull request):
```
public function getModuleName()
{
        $module = $this->argument('module');

        $module = app('modules')->find($module);

        if ($module === null) {
            return $module;
        }

        return $module->getStudlyName();
}
```
Can I apply my fix and made another pull request? :)

nWidart:
Oh I see ok.
In that case should the fix be assigning that call to a variable like you suggested?

huydang284:
Oh sorry @nWidart, you fix this issue in version 2.1.0
I am using Asgard and it still using laravel-modules version 2.0

nWidart:
Ah ok no problem. I'll close this pull request then. 😄

PR #549, opened 2018-06-12
Features/module factories

mohamed-aiman:
- added factory to config
- changed factories path from 
**project/Modules/ModuleName/Providers/Database/factories** to **project/Modules/ModuleName/Database/factories**
- changed **factories** to **Factories** for consistency, as by default other folders like Seeders,Migrations are named and others are named that way.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -99,6 +99,7 @@ return [
             'event' => 'Events',
             'listener' => 'Listeners',
             'migration' => 'Database/Migrations',
+            'factory' => 'Database/Factories',
             'model' => 'Entities',
             'repository' => 'Repositories',
             'seeder' => 'Database/Seeders',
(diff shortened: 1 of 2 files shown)


nWidart:
Thanks for the contribution!

Good catch for the location of factories!
From the diff I see `'factory' => 'Database/Factories',` was added, not changed, so it wasn't present before. Is this value used, or is it just as reference?

mohamed-aiman:
It's used and it works perfectly. while the module is generated and empty folder with keep is generated. I think its good to keep. :)

nWidart:
It looks like you're not up to date with latest changes. https://github.com/nWidart/laravel-modules/blob/master/config/config.php#L99 has a different structure.

nWidart:
And it also looks like the relative path issue was fixed too: https://github.com/nWidart/laravel-modules/blob/master/src/Commands/stubs/scaffold/provider.stub#L100

mohamed-aiman:
I have seen that. I think thats for the major version for php 7 and Larave >=5.5. I fixed it for version 1.0 which is used by php 5 > 7 and Laravel 5.4.

nWidart:
I see, 1.0 is not maintained anymore so you'll have to upgrade to version 2 to have this fix.

mohamed-aiman:
Cool 😊

PR #509, opened 2018-03-21
Returns the parameter `--class` to the `SeedCommand`.

danielsafronov:
This changes allows use `--class` option to specify a specific seeder class to run individually.

Example:
`php artisan module:seed User --class=Modules\User\Database\Seeders\UserRoleDatabaseSeeder`.

PR #591, opened 2018-07-24
fix module:seed --class

Instanced:
Example:
`php artisan module:seed User --class=UserSeeder`


Instanced:
@nWidart #509

nWidart:
And?

Instanced:
@nWidart This Pull requests mainly fixes the problem that `module:seed` doesn't work, and generates a class to module's seeds directory.

nWidart:
That still doesn't explain my question, but I'll try it out myself. 

Thanks for the contribution. 👍

Issue #465, opened 2018-01-19
php artisan module: seed Not working


githubmarau:
Hi, I've tried everything, the same error occurs.

"  In SeedCommand.php line 177:

  Type error: Argument 1 passed to Nwidart\Modules\Commands\SeedCommand::reportException() must be an instance of Exception, instance of Er
  ror given, called in F:\laragon\www\rhema\vendor\nwidart\laravel-modules\src\Commands\SeedCommand.php on line 50  "

<?php
namespace Modules\User\Database\Seeders;

use Illuminate\Database\Seeder;
use Illuminate\Database\Eloquent\Model;
use Modules\User\Database\Seeders\UserTableSeeder;

class UserDatabaseSeeder extends Seeder
{
    /**
     * Run the database seeds.
     *
     * @return void
     */
    public function run()
    {
        Model::unguard();

        //$this->call("UserTableSeeder");
        
        User::create([
            'name' => 'Admin',
            'email' => 'admin@admin.com',
            'password' => bcrypt(123456)
        ]);
    }
}

Could someone help me?

I've tried creating the UserTableSeeder class but the same error occurs.

raakesh:
I'm getting same error

`  Type error: Argument 1 passed to Nwidart\Modules\Commands\SeedCommand::reportException() must be an instance of Exception, instance of Error given, called in /home/vagrant/sites/perfqube/
  vendor/nwidart/laravel-modules/src/Commands/SeedCommand.php on line 50
`

raakesh:
Unable to seed Module based seeders.

andrietri:
im get same error

nWidart:
Feel free to send a pull request guys :)

iahmad:
for now I do it like so:

`php artisan db:seed --class=\\Modules\\Mymodule\\Database\\Seeders\\MymoduleDatabaseSeeder`

nWidart:
I was able to run the `module:seed` command with no issues here. How exactly could I reproduce the issue?

alisonvalente:
I had the same problem when following laravel documentation; then I realized I forgot to **use** the DB class. 
On UserDatabaseSeeder.php:

**use Illuminate\Support\Facades\DB;**

Works for me =)

h4ckedneko:
I think it's because `Error` does not inherit from `Exception` in PHP 7. But both of them inherit from `Throwable`, however it is not supported on PHP 5.

[...]

PR #598, opened 2018-08-01
Change throwable to Exception

filtration:
Laravel's exception handler requires Exception, but Throwable was passed

### DIFF
--- a/src/Commands/SeedCommand.php
+++ b/src/Commands/SeedCommand.php
@@ -46,7 +46,7 @@ class SeedCommand extends Command
                 array_walk($modules, [$this, 'moduleSeed']);
                 $this->info('All modules seeded.');
             }
-        } catch (\Throwable $e) {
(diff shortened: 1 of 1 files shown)


filtration:
When you seed a module, that relies on a class it will thrown an error, this pr fixed the issue of #465

nWidart:
I understand, however this was already tried, here : https://github.com/nWidart/laravel-modules/commit/cf27210c5079353dc65fb603f10b69a6a30ddf17#diff-25c026c4d45602db72947d8a0f1a2ab3

And was later changed what it is now. It never fixed the issue fully, for everyone.

pankajsanam:
I am also facing the same issue and changing Throwable to Exception in the handle method and reportException parameter is the only solution. There is no other way around.

nWidart:
That's not true no. There also reports when using Exception.

PR #621, opened 2018-09-03
Allow class resolution on short name and abstract

mikemand:
Allows developers to do: `app('modules')->someMethod();` and `app(Repository::class)->someMethod();`

Without this, there would be an exception thrown (for example in a new install of AsgardCMS v3):

```
Illuminate\Contracts\Container\BindingResolutionException thrown with message "Target [Nwidart\Modules\Repository] is not instantiable."
```

I'm also opening a PR to do the same thing for v3 of Laravel Modules.


mikemand:
v3 PR opened: #622

nWidart:
👍

PR #622, opened 2018-09-03
Allow class resolution on short name and abstract

mikemand:
Allows developers to do: `app(‘modules’)->someMethod();` and `app(FileRepository::class)->someMethod();`

This is the same as PR #621, but for v3 of Laravel Modules.

### DIFF
--- a/src/LaravelModulesServiceProvider.php
+++ b/src/LaravelModulesServiceProvider.php
@@ -44,10 +44,11 @@ class LaravelModulesServiceProvider extends ModulesServiceProvider
      */
     protected function registerServices()
     {
-        $this->app->singleton('modules', function ($app) {
+        $this->app->singleton(FileRepository::class, function ($app) {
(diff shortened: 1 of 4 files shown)


nWidart:
Sorry for the late reply but why not use the interface instead of the implementation? Ie https://github.com/nWidart/laravel-modules/blob/master/src/Contracts/RepositoryInterface.php

mikemand:
That's a good question. I'm not sure why the abstract was the one being used to bind to, instead of the interface. I will update both PRs. 😄

nWidart:
Actually, I just realised this would be a breaking change no?

Wouldn't it be better to do it on the next major version?

nWidart:
I don't understand how it's backwards compatible. You changed `modules` to the class name.

nWidart:
Thanks for the explanation, I'll add the extra tests.

PR #629, opened 2018-09-20
[5.7] disable mocking the console output & fix events

KABBOUCHI:
image

~There other test failures for the events, I'll check them later.~

I found the problem, while ago they changed the $event->fire to $event->dispatch, they remove it completely in 5.7


nWidart:
Wow, thank you! It's interesting that `withoutMockingConsoleOutput`  was the required change 😲

PR #633, opened 2018-09-22
API routes

arthvrian:
In Laravel 5.3+ in addition to web.php route file you also have api.php route file

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -22,7 +22,7 @@
     <logging>
         <log type="tap" target="build/report.tap"/>
         <log type="junit" target="build/report.junit.xml"/>
-        <log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>
+        <log type="coverage-html" target="build/coverage"/>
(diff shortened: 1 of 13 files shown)


phpunit.xml.dist
```diff
@@ -22,7 +22,7 @@
     <logging>
         <log type="tap" target="build/report.tap"/>
         <log type="junit" target="build/report.junit.xml"/>
-        <log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>
```
nWidart: Could you please add this back?

arthvrian: Done
I removed because of this
-----
Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 25:
  - Element 'log', attribute 'charset': The attribute 'charset' is not allowed.
  - Element 'log', attribute 'yui': The attribute 'yui' is not allowed.
  - Element 'log', attribute 'highlight': The attribute 'highlight' is not allowed.

  Test results may not be as expected.
-----

nWidart:
Thanks!

PR #662, opened 2018-11-11
update module controller scaffold to extend laravel default controller

naciriii:
Updating the module:make-controller command stub so that the created controller will extend App\Http\Controllers\Controller instead of use Illuminate\Routing\Controller.

### DIFF
--- a/src/Commands/stubs/controller-plain.stub
+++ b/src/Commands/stubs/controller-plain.stub
@@ -2,7 +2,8 @@
 
 namespace $CLASS_NAMESPACE$;
 
-use Illuminate\Routing\Controller;
+use Illuminate\Routing\Controller as BaseController;
+use App\Http\Controllers\Controller;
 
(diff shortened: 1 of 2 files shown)


nWidart:
Hello, thank you for the contribution.

What's the goal in this change? What problem does it fix?

(btw the tests are failing)

naciriii:
@nWidart  the goal of the change  is that the created controller will extend from Controller.php under App namespace so it can use validateRequests trait inside the controller, @KABBOUCHI  I see no reason to recreate a new base controller since it exists already, and the tests are already extending from Test/TestCase and this package is supposed to work along with laravel base scaffolding so I believe it's okay to extend Controller under App namspace

nWidart:
I think it's better to create your own base controller with those traits. In something like a core module. In my opinion, it's better to not depend on anything in the `app/` folder.

I understand this is an opinionated subject.

Issue #413, opened 2017-11-06
Unit Test per Module


Nightzer:
how can i make the unit test per module? i cant find it in the Artisan commands docs, so i make it manually per module.

nWidart:
Just like a package. There is nothing specific to laravel-modules for this.

spinyman:
I made a PR [#683](https://github.com/nWidart/laravel-modules/pull/683) for this.

PR #683, opened 2018-12-17
Test command

spinyman:
Due the issue https://github.com/nWidart/laravel-modules/issues/413


nWidart:
thank you for the contribution! However, can you please start your PR fresh from the current code base. The current PR has a lot of commits that aren't part of this PR which make it very hard to review.

PR #684, opened 2018-12-17
Test command

SpinyMan:
run command (module:test) for all modules or specified one.


spinyman:
Have you any problems with merge it?

nWidart:
Hello,

Sorry for the late reply. I've looked at your PR but I don't see a good reason to add this. It's already easy to use phpunit directly to run module tests.
You can view an [example phpunit config file here](https://github.com/AsgardCms/Platform/blob/3.0/phpunit.xml), test separate test suits are optional.

SpinyMan:
ok. if you think it's a best way - let it be.

nWidart:
Hi,

I'll close this to the before mentioned reason. I don't see a good use case for this over using phpunit itself.

PR #696, opened 2019-01-06
Add --api option to module:make-controller

kuptsovv:
Change controller stubs according to core laravel controller stubs


nWidart:

PR #697, opened 2019-01-08
add package to module by command

zedisdog:
now module can require package by command
```bash
php artisan module:require packageName moduleName
```
ps: This package is awesome.could I push this?I know the code seems....some.....stupid.:expressionless:


nWidart:
Hi and thank you for the contribution!

Could you please squash your commits so it makes just one commit please?

.gitignore
```diff
@@ -4,3 +4,4 @@ composer.lock
 /fixture/vendor
 build
 .php_cs.cache
+.idea
```
nWidart: Please remove this. This should go in your global gitignore file.

composer.json
```diff
@@ -18,7 +18,8 @@
     }
   ],
   "require": {
-    "php": ">=7.1"
+    "php": ">=7.1",
+    "ext-json": "*"
```
nWidart: Is this extension required by something we used?

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+     * @var string
+     */
+    protected $name = 'module:require';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'add dependencies for the specified module.';
```
nWidart: Please rename this to:
`Add a dependency for the specified module`

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        if (!$this->argument('packageName')) {
+            $this->error('packageName is required');
+            return;
+        }
+        $this->getModule();
```
nWidart: Why is this here? The return of the method isn't used.

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+
+    protected $signature = 'module:require {--dev} {packageName} {module?}';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        if (!$this->argument('packageName')) {
+            $this->error('packageName is required');
```
nWidart: Please rename the text to : `Argument packageName is required`

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+            $requireKey,
+            $requireOtherKey,
+            $newPackage,
+            $newPackageVersion,
+            $additionalScripts
+        );
+        $this->putComposer($moduleComposerPath, $moduleComposer);
+    }
+
+    protected function putComposer($path, $data)
```
nWidart: Could you rename this method to `writeComposerFile`?

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+            $this->info('package install success');
+            $this->info('updating module composer....');
+            $this->updateModuleComposer();
+            $this->info('updated module composer');
+        } else {
+            $this->warn('package install failed');
+        }
+    }
+
+    protected function getComposer($path)
```
nWidart: Could you rename this method to `getComposerFileContent`

nWidart: Also why is this method protected?

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * @param Json $moduleComposer
+     * @param string $requireKey
+     * @param string $requireOtherKey
+     * @param string $newPackage
+     * @param string $newPackageVersion
+     * @param array $additionalScripts
+     * @return Json
+     */
+    private function setModuleComposer($moduleComposer, $requireKey, $requireOtherKey, $newPackage, $newPackageVersion, $additionalScripts)
```
nWidart: Can you please rename this to `updateModuleComposerFile`

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+    protected function putComposer($path, $data)
+    {
+        file_put_contents($path, json_encode($data, JSON_UNESCAPED_SLASHES | JSON_PRETTY_PRINT));
+    }
+
+    /**
+     * @param array $composer
+     * @return array
+     */
+    private function getAdditional($composer)
```
nWidart: The method name is confusing to me. Get additional what? Can you be explicit here please?

src/Commands/RequireCommand.php
```diff
@@ -0,0 +1,153 @@
...
+        $newScripts = $composer['scripts'];
+        return array_diff_key($newScripts, $this->oldScripts);
+    }
+
+    /**
+     * @param array $composer
+     * @param string $require_key
+     * @return array
+     */
+    private function getNewPackageInfo($composer, $require_key)
```
nWidart: Please use `camelCase` as the parameter names.

nWidart:
Thank you for your work on this matter, however I'll have to close this PR for now.

I should probably document the recommended ways to handle this inside projects, based on how composer works.

zedisdog:
Hi,thanks for your time!I think there are some misunderstandings.

This command will first require package in project root folder,then copy package info from `composer.json in root folder` to `composer.json in the module folder`.

So the vendor folder won't be generated in module folder.The package which be added will only be placed in root vendor folder.

if you want it,I'm glad to submit a PR again with fixes which you mentioned above.:beers:

zedisdog:
Hi,by the way,I'm in china now.So you know,you work I sleep,you sleep I work.Please forgive me for my offline.:smile:

PR #698, opened 2019-01-09
Laravel Mix v4

KABBOUCHI:


nWidart:
Thanks for the contribution! 
It looks like you forgot to update the snapshot with: `-d --update-snapshots` to phpunit.

PR #713, opened 2019-02-05
Seeding modules not in Modules namespace.

daveismyname:
This adds the seeding ability to modules located in scans.paths config paths.

ie:
```
'scan' => [
    'enabled' => true,
    'paths' => [
        base_path('ModulesCms'),
    ],
],
```

The default running order remains the same if no modules are found in the Module namespace then any namespaces defined in the config will be tried.

### DIFF
--- a/src/Commands/SeedCommand.php
+++ b/src/Commands/SeedCommand.php
@@ -106,6 +106,14 @@ class SeedCommand extends Command
             $class = $this->getSeederName($name); //legacy support
             if (class_exists($class)) {
                 $seeders[] = $class;
+            } else {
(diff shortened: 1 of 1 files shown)


nWidart:

This is basically if a module is for instance in a vendor folder?

daveismyname:
No. but I guess that would apply to. My tests where when having another modules folder alongside the modules folder:

So in this case I have the normal Modules folder but also ModulesCms folder

screenshot 2019-02-24 at 13 36 10

nWidart:
Oh I see ok. 👍

PR #724, opened 2019-03-04
Update code for Laravel 5.8 release.

Schnoop:
Adopted changes that has been made to Laravel in 5.8 release.

### DIFF
--- a/src/ModulesServiceProvider.php
+++ b/src/ModulesServiceProvider.php
@@ -9,12 +9,6 @@ use Nwidart\Modules\Providers\ContractsServiceProvider;
 
 abstract class ModulesServiceProvider extends ServiceProvider
 {
-    /**
-     * Indicates if loading of the provider is deferred.
-     *
-     * @var bool
-     */
-    protected $defer = false;
 
     /**
      * Booting the package.
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 | 5.5  | ^2.0  |
 | 5.6  | ^3.0  |
 | 5.7  | ^4.0  |
+| 5.8  | ^5.0  |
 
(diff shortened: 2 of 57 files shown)


nWidart:
Hello,

I just pushed the 5.8 update too at the same time :P 
https://github.com/nWidart/laravel-modules on master branch

Schnoop:
:D Bad luck. I've also adopted that changes made to the string and array helper methods and therefore would like to create a new release 5.0.
Also there are is a depreciation in the $defer car in the Service Providers.

src/ModulesServiceProvider.php
```diff
@@ -9,12 +9,6 @@
 
 abstract class ModulesServiceProvider extends ServiceProvider
 {
-    /**
-     * Indicates if loading of the provider is deferred.
-     *
-     * @var bool
-     */
-    protected $defer = false;
```
nWidart: Was this removed in laravel?

Schnoop: > Deferred Service Providers
Likelihood Of Impact: Medium

The defer boolean property on the service provider which is/was used to indicate if a provider is deferred has been deprecated. In order to mark the service provider as deferred it should implement the Illuminate\Contracts\Support\DeferrableProvider contract.

nWidart:
Thanks for the depreciation changes, applied those on master branch 👍

src/ModulesServiceProvider.php
nWidart: Oh nice ok.

Schnoop:
Why did you drop PHP 7.1 support? Because of this, i've to stick to Laravel 5.7 :(

nWidart:
Because phpunit 8 doesn't support 7.1 anymore.

Schnoop:
There's no need to use PHPUnit 8. Laravel itself sticks to 7.

nWidart:
Yes but I would like to use up to date packages 😄

PR #727, opened 2019-03-07
Update Route Provider stub to not affect the root namespace of the URL generator

mfullbrook:
The stubbed RouteProvider extends `Illuminate\Foundation\Support\Providers\RouteServiceProvider` setting the `$namespace` property means that on each `boot()` of a module's RouteServiceProvider we are changing the URL generator's root namespace.  

I believe this should be controlled by the application level RouteServiceProvider only.
See `Illuminate\Foundation\Support\ProvidersRouteServiceProvider::setRootControllerNamespace`


nWidart:
Thanks this makes sense!

PR #733, opened 2019-03-15
Changed SeedCommand to use \Exception instead of \Throawble

### LINKED ISSUE #574: ExceptionHandler::report() mu st be an instance of Exception

 179"

Thanks for your help!

Rezrazi:
Fixes #574

### DIFF
--- a/src/Commands/SeedCommand.php
+++ b/src/Commands/SeedCommand.php
@@ -46,7 +46,7 @@ class SeedCommand extends Command
                 array_walk($modules, [$this, 'moduleSeed']);
                 $this->info('All modules seeded.');
             }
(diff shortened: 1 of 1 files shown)


nWidart:
Hi and thank you for the contribution. 

However, as mentioned: https://github.com/nWidart/laravel-modules/issues/574#issuecomment-412005995

This will not fix the issue.

Rezrazi:
Yes indeed. I should've looked at previous PRs before this one.
Could you share the environment you're using in which case the original code works ?
I see this as weird as it might work on in some situations and not in some others, as it seems to be a Laravel related issue.

For my case it's 
- Laravel 5.8
- PHP 7.3
- Package version 4.1

PR #744, opened 2019-04-02
Convert Error to ErrorException on SeedCommand

### LINKED ISSUE #574: ExceptionHandler::report() mu st be an instance of Exception

stephandesouza:
Fixes \Throwable mess in Laravel report.

Fixes #574 using part of #598. 

Is the same tecnique used by [\Psysh\ExecutionLoopClosure:95](https://github.com/bobthecow/psysh/blob/master/src/ExecutionLoopClosure.php#L95)

Now instead of this:
Annotation 2019-04-02 092930

We got this:
Annotation 2019-04-02 092822

### DIFF
--- a/src/Commands/SeedCommand.php
+++ b/src/Commands/SeedCommand.php
@@ -46,7 +46,14 @@ class SeedCommand extends Command
                 array_walk($modules, [$this, 'moduleSeed']);
                 $this->info('All modules seeded.');
             }
-        } catch (\Throwable $e) {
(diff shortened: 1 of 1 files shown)


nWidart:
Hi, thank you for the contribution but this is not a fix.
This was already mentioned in other tickets and other PRs

ikidnapmyself:
I think it's not the most elegant way to fix it, but that by passes this issue. Basically returns the exact error message.

PR #748, opened 2019-04-06
Refactor TestMakeCommand / add `--feature` flag

artemsky:
- Refactor `unit-test` stub to be consistent with Laravel 5.8
- Add `feature-test.stub`
- Change namespaces according to Laravel 5.8

### WARNING - Changes current behavior

Before:
- `php artisan module:make-test EloquentPostRepositoryTest Blog` created a test in `Test` namespace

Now:
- `php artisan module:make-test EloquentPostRepositoryTest Blog --feature` create a test in `Test\Feature` namespace
- `php artisan module:make-test EloquentPostRepositoryTest Blog` create a test in `Test\Unit` namespace

Docs update - https://github.com/nWidart/nwidart-docs/pull/21


artemsky:
In some reasons, can't run tests on my local env. Will fix it in a few days

nWidart:
Hi and thank you for the contribution!

I think it would be better to make the unit test default and the feature one the additional flag.

src/Commands/TestMakeCommand.php
```diff
@@ -18,7 +19,10 @@ class TestMakeCommand extends GeneratorCommand
 
     public function getDefaultNamespace() : string
     {
-        return $this->laravel['modules']->config('paths.generator.test.path', 'Tests');
+        if ($this->option('unit')) {
```
nWidart: This looks like unit is not default yet, shouldn't this be changed to feature too?

src/Commands/TestMakeCommand.php
```diff
@@ -34,14 +38,30 @@ protected function getArguments()
...
 
+    /**
+     * Get the console command options.
+     *
+     * @return array
+     */
+    protected function getOptions()
+    {
+        return [
+            ['unit', false, InputOption::VALUE_NONE, 'Create a unit test.'],
```
nWidart: Shouldn't this be feature?

config/config.php
```diff
@@ -110,7 +110,8 @@
             'assets' => ['path' => 'Resources/assets', 'generate' => true],
             'lang' => ['path' => 'Resources/lang', 'generate' => true],
             'views' => ['path' => 'Resources/views', 'generate' => true],
-            'test' => ['path' => 'Tests', 'generate' => true],
+            'test-feature' => ['path' => 'Tests/Feature', 'generate' => true],
+            'test-unit' => ['path' => 'Tests/Unit', 'generate' => true],
```
nWidart: This looks like a breaking change. We need to keep the default `test` key to not break backwards compatibility.

artemsky: ok

nWidart:
Thanks for the changes! 👍

PR #758, opened 2019-05-12
[WIP] Using laravel's native migrate commands

nWidart:
This is a work in progress pull request to replace the custom `Migrator` class by using laravel's native `migrate:` commands.
This is an experiment and is not guaranteed to merged.

Current module command would still exist, but as a proxy to the native command.

| laravel-modules  | native  | status |
|---|---|---|
| `module:migrate`  | `migrate`  | ✅|
| `module:migrate-reset` | `migrate:reset`  | in progress |
| `module:migrate-rollback` | `migrate:rollback`  | in progress |
| `module:migrate-status` | `migrate:status`  |  |

***

### `migrate:reset`

When we have more than 1 module, the migration table will contain migrations for multiple modules:

image

Knowing this, when running `module:reset` (proxying to `migrate:reset`), it loops over every module, and calls the reset command.
This will show:

```
╰─ art module:migrate-reset
Running for module: Blog
Migration not found: 2019_03_04_121211_create_recipes_table
Rolling back: 2019_05_12_160047_create_posts_table
Rolled back:  2019_05_12_160047_create_posts_table
Running for module: Recipe
Rolling back: 2019_03_04_121211_create_recipes_table
Rolled back:  2019_03_04_121211_create_recipes_table
```

It shows a "migration not found" error, because when trying to reset the first module, `Blog`, it sees there's another migration name in the database, but cannot find the file associated with it.

This will probably be recurring in other commands like rollback.
I'm currently not sure how to fix this.

Fix when calling for all modules: Instead of looping over every module and calling `reset`, prepare the array of paths and call reset with this.
Issue remains when calling reset on a specific module.

### `migrate:rollback`

Unlike `migrate:reset`, `migrate:rollback` does not support getting an array of paths.
Depends on: https://github.com/laravel/framework/pull/28495

### `migrate:status`

Depends on: https://github.com/laravel/framework/pull/28496


narwy:
@nWidart you got any issues migrating other modules, seem they don't drop the database and it's contents and just throw errors that it's not empty 🤷‍♂️

nWidart:
You mean the message mentioned in the original post:
```
Migration not found: 2019_03_04_121211_create_recipes_table
```
?

narwy:
@nWidart I seem to have issues where the `down()` function in any of my migrations in modules isn't registering. It's not hitting that function, latest version and all. Are you aware of this? I have to drop the database manually (another connection) for it to work 🤷‍♂️

```
<?php

use Illuminate\Support\Facades\Schema;
use Illuminate\Database\Migrations\Migration;

class CreateBatchesTable extends Migration
{
    /**
     * Run the migrations.
     */
    public function up()
    {
        Schema::connection('funder')->create('batches');
    }

    /**
     * Reverse the migrations.
     */
    public function down()
    {
        Schema::connection('funder')->drop('batches');
    }
}
```

nWidart:
Hm not sure about that, I would assume so?

narwy:
> Hm not sure about that, I would assume so?

Can you give me an example how to use the force command, can't seem to get it working on any of the laravel-modules commands, nothing is doing the `drop()` function in the migration

nWidart:
I doubt it's relevant in the PR 😄 
But you can run the module migrate command multiple times with `--force` for example.

narwy:
> I doubt it's relevant in the PR 😄
> But you can run the module migrate command multiple times with `--force` for example.

Yeh the `--force` doesn't do anything.

rcbgalido:
@nWidart I think one thing you could is trigger ```migrate:reset``` using ```Artisan::call()``` and retrieve the output with ```Artisan::output()```. From there, you could manipulate the output by removing all lines that starts with ```Migration not found```.

rcbgalido:
@nWidart I don't know if you're still planning to finish this feature but one thing you could do is to execute ```migrate:reset``` using ```Artisan::call()``` and retrieve the output by ```Artisan::output()```. From there, you could manipulate the output by removing the lines that starts with ```Migration not found```.

PR #759, opened 2019-05-13
Fix exception issue by replacing throwable with built-in php exception

### LINKED ISSUE #574: ExceptionHandler::report() mu st be an instance of Exception

nerdial:
This PR fixes  #574 issue . 
**The problem is that, laravel expects to get php built-in \Exception but that method gives it an instance of \Throwable interface which is wrong in php concept. because even \Exception is instance of \Throwable but the opposite is not possible. so the workaround is to pass an \Exception to it.** 
this really fixed my issues in  src/Commands/SeedCommand.php , i hope this pr fixes yours too. 
i checked previous issues about this problem, feel free to report other problems here , i will gladly fix them too. by the way i ran all of unit tests in the code and everything was fine.

### DIFF
--- a/src/Commands/SeedCommand.php
+++ b/src/Commands/SeedCommand.php
@@ -46,7 +46,7 @@ class SeedCommand extends Command
                 array_walk($modules, [$this, 'moduleSeed']);
                 $this->info('All modules seeded.');
             }
-        } catch (\Throwable $e) {
+        } catch (\Exception $e) {
             $this->reportException($e);
 
             $this->renderException($this->getOutput(), $e);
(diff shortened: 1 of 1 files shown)


nWidart:
Thanks for you contribution.

This was mentioned many times before, it does not fix the issue. Both can be thrown.

nerdial:
@nWidart what is the issue ?

nWidart:
Look at many previous pull requests trying to fix the issue, as well as the thread you linked.


Maybe the solution could be to catch both.

nerdial:
i looked at other issues most of them was solved by replacing throwable with exception .
what other problems are there ? why this fix can't resolve the issue ?

nWidart:
"most", I'm telling you that others will have the same issue, but for throwable.

PR #773, opened 2019-06-12
Only Register Factories when Running in Console

nmfzone:
I think this approach can reduce the application to load the unused file when running in browser.


nWidart:
Thanks, good idea! 👌

nmfzone:
Sure 👌
But, I'm not sure why Travis CI fails. I think you should update the tests snapshot.

nWidart:
Indeed, that's something you need to run on your end.

nmfzone:
@nWidart Ohh, I see..on my way.

nWidart:
Thanks!

PR #776, opened 2019-06-19
Add alternative way to provide the namespace via configuration

daison12006013:
We could add a third option inside **paths.generator.{alias}** to have `namespace`

When adding parent folder, as part of the `path`, the current behaviour uses the `path` as well as the namespace, there is no way for us to easily provide the namespace, or else we are REQUIRED to extends the commands and modify the laravel module's ConsoleServiceProvider and the booted Provider.

What I could suggest is to add a 3rd parameter as `optional`, so there is no breaking change @nWidart 

```
[
    'path' => 'src/Providers',
    'generate' => true,
    'namespace' => 'Providers',
]
```

---

Example modules that I have and their structure

image

### DIFF
--- a/src/Commands/CommandMakeCommand.php
+++ b/src/Commands/CommandMakeCommand.php
@@ -38,9 +38,7 @@ class CommandMakeCommand extends GeneratorCommand
     {
         $module = $this->laravel['modules'];
 
-        return ($namespace = $module->config('paths.generator.command.namespace'))
-            ? $namespace
-            : $module->config('paths.generator.command.path', 'Console');
+        return $module->config('paths.generator.command.namespace') ?: $module->config('paths.generator.command.path', 'Console');
     }
 
     /**
--- a/src/Commands/CommandMakeCommand.php
+++ b/src/Commands/CommandMakeCommand.php
@@ -36,7 +36,11 @@ class CommandMakeCommand extends GeneratorCommand
 
     public function getDefaultNamespace() : string
     {
-        return $this->laravel['modules']->config('paths.generator.command.path', 'Console');
+        $module = $this->laravel['modules'];
+
+        return ($namespace = $module->config('paths.generator.command.namespace'))
+            ? $namespace
+            : $module->config('paths.generator.command.path', 'Console');
     }
 
     /**
--- a/src/Commands/ControllerMakeCommand.php
+++ b/src/Commands/ControllerMakeCommand.php
@@ -119,9 +119,7 @@ class ControllerMakeCommand extends GeneratorCommand
     {
         $module = $this->laravel['modules'];
 
-        return ($namespace = $module->config('paths.generator.controller.namespace'))
-            ? $namespace
-            : $module->config('paths.generator.controller.path', 'Http/Controllers');
+        return $module->config('paths.generator.controller.namespace') ?: $module->config('paths.generator.controller.path', 'Http/Controllers');
     }
 
     /**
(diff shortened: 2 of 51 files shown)


nWidart:
Hi,

Thanks for your contribution! However, I don't see a use-case for this. In php the namespace should always (or is recommended) match the folder structure. 
Via the configuration, you can already override the default namespace used.

daison12006013:
@nWidart, there is no way for us to modify the path, your console commands were limited because of their default namespace.

You're basically using the `path` as their namespace.

Sometimes creating a module or a package you wrapped things up inside an `src/` folder.

E.g, generate a module called `PhotoProcessor` If you will try to go inside `PhotoProceessorServiceProvider.php` the namespace will always be like this.

```
<?php

namespace Modules\src\PhotoProcessor\Providers;
```

Please try yourself.

nWidart:
So this is a different issue? The src subfolder is something else, that was addressed in previous issues.

We shouldn't use the namespace to dictate the folder structure if we can do it the other way around.
The paths of each generated file can be configured via [the configuration](https://github.com/nWidart/laravel-modules/blob/master/config/config.php#L99).

daison12006013:
> The src subfolder is something else, that was addressed in previous issues.

So any solutions on this? I'm not referring to tha path, but the namespace itself, do we have solutions on this?

```
<?php

namespace Modules\src\PhotoProcessor\Providers;
```

nWidart:
Not really, just recommended to not do it.

Though if this PR enables this feature, why not add it indeed, I'll test it out in a bit.

nWidart:
Tested it and seems to work very well. 👌 
I'll have a couple of comments on the code itself.

About the autoloading do you autoload every module manually, or is there a way to have one autoload line for all modules.

```json
"Modules\\Blog\\": "Modules/Blog/src"
```

src/Commands/CommandMakeCommand.php
```diff
@@ -36,7 +36,11 @@ class CommandMakeCommand extends GeneratorCommand
 
     public function getDefaultNamespace() : string
     {
-        return $this->laravel['modules']->config('paths.generator.command.path', 'Console');
+        $module = $this->laravel['modules'];
+
+        return ($namespace = $module->config('paths.generator.command.namespace'))
```
nWidart: Why catching this in a `namespace` variable? I think you can use `?:` instead. In order to have it fit on one line this can be done:

```php
    public function getDefaultNamespace() : string
    {
        $module = $this->laravel['modules'];

        $namespace = $module->config('paths.generator.command.namespace');
        
        return $namespace ?: $module->config('paths.generator.command.path', 'Console');
    }
```
(same for all)

daison12006013: Copy will add unit and some changes on it

daison12006013:
> About the autoloading do you autoload every module manually, or is there a way to have one autoload line for all modules.
> 
> ```json
> "Modules\\Blog\\": "Modules/Blog/src"
> ```

Should I add another config (by default it is empty) and if developers would like to modify the value into "src/" then inside `composer.stub` we could pull the value.

nWidart:
I didn't understand your last comment, you're updating your PR?

daison12006013:
> I didn't understand your last comment, you're updating your PR?

Forget my last comment, I am just thinking a different approach for the autoload path.
Will update my PR soon, quite busy yet.

daison12006013:
@nWidart updated

aimensasi:
any updates on the autoload issue, cause right now I had to do this https://github.com/nWidart/laravel-modules/pull/776#issuecomment-503948634 for every module

daison12006013:
@aimensasi there are things that limits us, since we're using stub/template thing, might be better to request a new feature for it referencing this Pull Request? so we could focus into that issue.

Issue #789, opened 2019-07-07
Activator feature


ryssbowh:
Hi there,

In my project I needed to take the activation logic out of the Module class, the main reason for this being that the activation status being in a git tracked file, I find it makes it messy when pushing changes to modules (as composer packages) and pulling them in other projects.
If I'm developing a module, it's obviously enabled, but I might not want it enabled by default when pulling it in another project, maybe I have some logic to be made before etc..

Anyway I've implemented that in my project, with a default FileActivator that stores the activation status in a separate file, but it could be a database activator or anything else.

I'm not doing a pull request as it's quite a change but if that's of any interest I'll do one https://github.com/ryssbowh/laravel-modules/tree/activator/.

Thanks again for this great package !

nWidart:
You can take a look at the [database-modules[(https://github.com/nWidart/laravel-modules/tree/feature/database-modules) branch for this too.

nWidart:
I just looked at your branch, I like the idea of this. 👌  It's a good intermediary solution.

ryssbowh:
Thanks,

I did have a look at the database branch, but didn't feel I had the level to take it on.

Doing a pull request now.

SeoRoman:
@nWidart What all does the database-modules system still need?

lexdewilligen:
@nWidart I've started working on the database implementation using your database branch. I've implemented some methods that were abstract, but I don't know how to go on. Could you take a look at my branch and tell me what needs to be done next? 

https://github.com/lexdewilligen/laravel-modules/commit/0fe036eb7f81d904b641f4780058bd3431a0c2b1

nWidart:
Thank you!

I believe if you need the database implementation for maintaining module statuses, using the activator feature will be your fastest way to get this up and running.

lexdewilligen:
@nWidart Yes, but I don't understand how to implement the activator feature for the database..

ryssbowh:
Hi,

If you write a new Activator class that implements the ActivatorInterface and reference it in your config :
```
'activators' => [
        'file' => [
            'class' => FileActivator::class,
            'statuses-file' => storage_path('modules_statuses'),
            'cache-key' => 'activator.installed',
            'cache-lifetime' => 604800,
        ],
        'database' => [
            'class' => DatabaseActivator::class
        ]
    ],

    'activator' => 'database'
```
That should work

lexdewilligen:
[...]

PR #790, opened 2019-07-07
Activator

ryssbowh:
Hi,

as discussed here : https://github.com/nWidart/laravel-modules/issues/789


config/config.php
```diff
@@ -166,7 +168,7 @@
     'cache' => [
         'enabled' => false,
         'key' => 'laravel-modules',
-        'lifetime' => 60,
+        'lifetime' => 60
```
mikemand: Why? Trailing commas is pretty much a PSR requirement, which this package follows.

config/config.php
```diff
@@ -186,4 +188,24 @@
...
+    | You can define new types of activators here, file, database etc. The only
+    | required parameter is 'class'.
+    | The file activator will store the activation status in storage/installed_modules
+    */
+    'activators' => [
+        'file' => [
+            'class' => FileActivator::class,
+            'file' => storage_path('installed_modules'),
+            'cache-key' => 'activator.installed',
+            'cache-lifetime' => 604800
```
mikemand: Needs trailing commas.

ryssbowh:
Yes, the cache lifetime for the file activator is in seconds.

Trailing commas... don't know when my brain will get used to that :)

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+		$this->files = $app['files'];
+		$this->config = $app['config'];
+		$this->fileInstalled = $this->config('file');
+		$this->cacheKey = $this->config('cache-key');
+		$this->cacheLifetime = $this->config('cache-lifetime');
+		$this->installed = $this->getCached();
+	}
+
+	/**
+	 * Get installed cache
```
nWidart: This comment looks a bit unclear to me. From the comment alone, I expected this to return laravel's cache implementation.

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @var string
+	 */
+	protected $cacheLifetime;
+
+	/**
+	 * Array of modules activation statuses
+	 * 
+	 * @var array
+	 */
+	protected $installed;
```
nWidart: Could this be renamed to `$installedModules`. I was hesitating if this was a boolean or not. 

Also what's an "installed module"? Is it an active module? If so, it might be more clear to use `activeModules` instead.

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+use Illuminate\Cache\CacheManager;
+use Illuminate\Config\Repository as Config;
+use Illuminate\Container\Container;
+use Illuminate\Filesystem\Filesystem;
+use Nwidart\Modules\Contracts\ActivatorInterface;
+use Nwidart\Modules\Module;
+
+class FileActivator implements ActivatorInterface
+{
+	/**
```
nWidart: Why are fields/methods protected instead of private? Could you make all `protected` fields and methods private?

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+		$this->cacheLifetime = $this->config('cache-lifetime');
+		$this->installed = $this->getCached();
+	}
+
+	/**
+	 * Get installed cache
+	 * 
+	 * @return array
+	 */
+	protected function getCached()
```
nWidart: Maybe the method could be named `getCachedModules`?

It seems like this doesn't always returns cached modules. If the cache is disabled, it won't from the looks of it, which doesn't seem to be the responsibility of this method, based on its name.

Actually, the only usage of this method is the constructor, so maybe this shouldn't mention cached at all and just be called `getInstalledModules` or `getActiveModules`.

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+	{
+		if(!$this->config->get('modules.cache.enabled')) return $this->readJson();
+		
+		return $this->cache->remember($this->cacheKey, $this->cacheLifetime, function () {
+            return $this->readJson();
+        });
+	}
+
+	/**
+	 * Forgets the installed cache
```
nWidart: Comment seems weird to me, we don't delete cache, we delete the contents of the cache for a specific key in this implementation. 
I would call this `flushCache` or `clearCache`.

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @var string
+	 */
+	protected $fileInstalled;
+
+	public function __construct(Container $app)
+	{
+		$this->cache = $app['cache'];
+		$this->files = $app['files'];
+		$this->config = $app['config'];
+		$this->fileInstalled = $this->config('file');
```
nWidart: `fileInstalled` seems weird. Also the fact that there's a difference between the config key and the class field means there's something off with the naming. 
It should be the same in config key and class field. 

In order to chose a good name, What is this file? What does it contain? 
It looks like it contains a list of all active modules? 
If so we should use that keyword and avoid "install"

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @param  $default
+	 * @return mixed
+	 */
+	protected function config(string $key, $default = null)
+    {
+        return $this->config->get('modules.activators.file.' . $key, $default);
+    }
+
+	/**
+	 * Reads the installed json file
```
nWidart: Can a json file be uninstalled? I think this read the json containing the active modules?

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @return array
+	 */
+	protected function readJson()
+	{
+		if(!$this->files->exists($this->fileInstalled)) return [];
+		return json_decode($this->files->get($this->fileInstalled), true);
+	}
+
+	/**
+	 * Writes the installed json file
```
nWidart: Same here, I wouldn't use "install"

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+		$this->setActiveByName($module->getName(), 0);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isStatus(Module $module, $status)
+    {
+    	if(!isset($this->installed[$module->getName()])) return false;
+    	return $this->installed[$module->getName()] == $status;
```
nWidart: Could you use strict comparison here, please? `===`

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public function disable(Module $module)
+    {
+		$this->setActiveByName($module->getName(), 0);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isStatus(Module $module, $status)
```
nWidart: What's the type of the second argument, let's typehint this, I assume it's a boolean.

src/Activators/FileActivator.php
```diff
@@ -0,0 +1,175 @@
...
+    {
+    	$this->installed[$name] = (int)$status;
+    	$this->writeJson();
+    	$this->forgetCache();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function delete(Module $module)
```
nWidart: This doesn't delete the module right? So let's call it `disableModule`?
Actually, I see there's also a disable method? So why would we ever want to delete it?

src/Commands/ModuleMakeCommand.php
```diff
@@ -35,9 +35,11 @@ public function handle()
                 ->setFilesystem($this->laravel['files'])
                 ->setModule($this->laravel['modules'])
                 ->setConfig($this->laravel['config'])
+                ->setActivator($this->laravel['modules.activator'])
```
nWidart: Let's try to avoid the alias key and instead use the class reference. I'm trying to move away from the aliases.

src/Contracts/ActivatorInterface.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param  $active
+     */
+    public function setActiveByName(string $name, $active);
+
+    /**
+     * Deletes a module activation status
+     * 
+     * @param  Module $module
+     */
+    public function delete(Module $module);
```
nWidart: There's delete and disable? What's the difference?

src/Generators/ModuleGenerator.php
```diff
@@ -97,6 +114,20 @@ public function setPlain($plain)
...
     }
 
+    /**
+     * Set active flag.
+     *
+     * @param bool $active
+     *
+     * @return $this
+     */
+    public function setActive($active)
```
nWidart: Let's typehint the argument to boolean?

src/Generators/ModuleGenerator.php
```diff
@@ -97,6 +114,20 @@ public function setPlain($plain)
...
+    {
+        $this->active = $active;
```
nWidart: Since this is a boolean, please rename to `isActive`.

src/Generators/ModuleGenerator.php
```diff
@@ -131,6 +162,20 @@ public function setConfig($config)
...
     }
 
+    /**
+     * Set the modules activator
+     *
+     * @param ActivatorInterface $activator
+     *
+     * @return $this
+     */
+    public function setActivator($activator)
```
nWidart: You can typehint the interface here too.

src/LaravelModulesServiceProvider.php
```diff
@@ -52,6 +52,13 @@ protected function registerServices()
...
             return new Laravel\LaravelFileRepository($app, $path);
         });
+        $this->app->singleton(Contracts\ActivatorInterface::class, function ($app) {
+            $activator = $app['config']->get('modules.activator');
+            $class = $app['config']->get('modules.activators.' . $activator)['class'];
+
+            return new $class($app);
+        });
         $this->app->alias(Contracts\RepositoryInterface::class, 'modules');
+        $this->app->alias(Contracts\ActivatorInterface::class, 'modules.activator');
```
nWidart: I wouldn't create the alias here, I'm trying to move away from this practice.

nWidart:
This is a breaking change so it'll have to be for the next major release.

nWidart:
It looks like there's no tests at all for this new FileActivate class, this should be added.

src/Generators/ModuleGenerator.php
ryssbowh: I'd agree. But this argument comes from the Module class and it's not type hinted there, should I type hint it there too ?

src/Contracts/ActivatorInterface.php
ryssbowh: This was only to make up for the delete method of the Module class. I assumed it's just a helper for developers or tests since I don't see any reference of it (?).
It basically deletes any activation status for that module (active or not)

src/Activators/FileActivator.php
ryssbowh: The way I see it, an installed module is a module that exists on the disk, it then has an activation status (1 or 0), both are stored by this activator.

nWidart: Hm, thanks that explains the keyword a bit. So this contains an array of all modules? If so I'd still rename it `installedModules` to be explicit.

src/Contracts/ActivatorInterface.php
nWidart: I see ok, that makes sense. I believe the delete method on the module class is there by request from the community, there could a future delete command that uses that too actually.

src/Generators/ModuleGenerator.php
nWidart: Hm, this can still be type hinted without changing the Module class no? I think that would be a first step.
This package existed before scalar type hints in php so I'm only now moving towards them over time.

ryssbowh: Ok, to be consistent with the Module class, I'll have to type hint them as integers even though they can only be 0 or 1. And they will have to be changed later, is that what you mean ?

nWidart: Indeed, now that I think of it, maybe there's no need to have them as int, as there's no other status other than 0 or 1. Meaning the could be booleans.
If possible you can typehint on both sides, if not, don't worry I'll do that later on myself.

ryssbowh:
Quite struggling with the tests as many tests (FileRepository, Module...) change module's activation statuses now, making the next tests assertions fail as the statuses are written in a separate file that is common to all tests, what would be the way to go you think :

- Each test that by its actions create a statuses file should delete it when it ends ?
- Or the following tests should take into account the previous tests activation statuses ?
- something else ?

nWidart:
Each test should be independent of one another so I think deleting the file after each test should be good.

ryssbowh:
A few things about the last 2 commits : 

- removed tests/stubs/valid/DisabledModule as status is not stored within the module folder anymore
- removed all statuses within tests/*/module.json

- I've not tested all the Lumen parts of these commits

- added a reset method to ActivatorInterface as I needed to clear data after the tests
- changed all statuses typehints to bools

nWidart:
Thank you for your patience.

nWidart:
We should document this change in the docs, I'm confused now how to activate a module.

nWidart:
It looks like we need a command to migrate from the old way to handle active status, and write them in the activator file.

ryssbowh:
The way to activate a module hasn't changed, through the Module\enable() and \Module\disable().

Good point with the command, I'll write this asap.

nWidart:
I see I was wondering because it didn't detect old modules but that's indeed because they're not yet added to the new file.

I'm working on the migration command currently.

nWidart:
The current command can be viewed here: https://github.com/nWidart/laravel-modules/commit/a70af0dc04e26cc4f34bccdb08bb30591c9e155a

Let me know if I forgot something.

ryssbowh:
The only thing I can think of is that this command will fire 'enabling' and 'enable' events for each module, which you might not want to happen given that the modules are not technically enabled.

Maybe you can fetch the file used by the activator `$activator->getStatusesFilePath()` and write a statuses array 
```
[
  "Core" => true,
  "Media" => false
]
```
as json inside it.

And clear the activator cache afterwards. The key is `config('modules.activators.file.cache-key')`

ryssbowh:
Also,

It might be better to do this work within the activator class, in case someone writes a database handled activator that would need a different fix to this problem

nWidart:
Hm not sure if the events being fired is an actual issue. Since the events for enabled/disabled weren't really used before I doubt they do anything critical in user's codebases.

I'm not sure about doing this in the activator class, as it will have to know from which version to which version which probably isn't his responsibility. Maybe another class though.

PR #807, opened 2019-08-18
Fix namespace generation

Davidnadejdin:


nWidart:

You have to update the test snapshots for the build to succeed.

PR #830, opened 2019-09-12
Change the fileactivator status file path

RSpeekenbrink:
Change the default path for the statuses file for the FileActivator to the `app` folder in the storage folder. This to keep the application's storage folder clean.

Point of discussion; maybe add modules folder in storage and include files like this in there? The files app folder get git ignored by default.


nWidart:

PR #857, opened 2019-10-17
Feature/database-activator

lexdewilligen:
Hi! This PR is a working implementation of the database activator, however it needs quite a few improvements. I don't have enough time to improve the database implementation, however I need this part in some projects. I'd therefore like this PR to be merged already so other can also improve the database implementation.

### DIFF
--- a/src/Commands/EnableCommand.php
+++ b/src/Commands/EnableCommand.php
@@ -28,7 +28,7 @@ class EnableCommand extends Command
     public function handle()
     {
         /** @var Module $module */
-        $module = $this->laravel['modules']->findOrFail($this->argument('module'));
+        $module = $this->laravel['modules']->findByNameOrCreate($this->argument('module'));
 
         if ($module->isDisabled()) {
             $module->enable();
--- a/src/LaravelModulesServiceProvider.php
+++ b/src/LaravelModulesServiceProvider.php
@@ -4,7 +4,6 @@ namespace Nwidart\Modules;
 
 use Illuminate\Support\Facades\Schema;
 use Nwidart\Modules\Contracts\RepositoryInterface;
-use Nwidart\Modules\Exceptions\ModulesTableNotFoundException;
 use Nwidart\Modules\Support\Stub;
 
 class LaravelModulesServiceProvider extends ModulesServiceProvider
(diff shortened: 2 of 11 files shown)


nWidart:
Hi,

Thanks for your contribution! 
I can't merge this yet because of:
- failing tests
- missing tests for the new feature.

You can however already use this in your project, you don't have to have your code merged.
https://stackoverflow.com/questions/13498519/how-to-require-a-fork-with-composer

src/LaravelModulesServiceProvider.php
```diff
@@ -12,6 +13,10 @@ class LaravelModulesServiceProvider extends ModulesServiceProvider
      */
     public function boot()
     {
+        if (!Schema::hasTable(config('modules.activators.database.table'))) {
```
nWidart: This can't be here.

lexdewilligen: It has to be since you will get an error that the database does not have a table 'modules' because the provider is booted before you can migrate.

nWidart: This is the reason all tests fail. This is also related to the implementation in a place where implementation can't be referenced.

src/Commands/EnableCommand.php
```diff
@@ -28,7 +28,7 @@ class EnableCommand extends Command
     public function handle()
     {
         /** @var Module $module */
-        $module = $this->laravel['modules']->findOrFail($this->argument('module'));
+        $module = $this->laravel['modules']->findByNameOrCreate($this->argument('module'));
```
nWidart: Don't use this, or implement it in all implementations.

src/Module.php
```diff
@@ -203,7 +203,7 @@ protected function registerTranslation(): void
     {
         $lowerName = $this->getLowerName();
 
-        $langPath = $this->getPath() . '/Resources/lang';
+        $langPath = $this->getPath() . config('modules.paths.generator.lang.namespace', config('modules.paths.generator.lang.path'));
```
nWidart: This makes tests fail too. Looking into it.

peter-mw:
great feature, looking forward to use it

narwy:
I agree with @nWidart the `LaravelModulesServiceProvider` can't have that kind of if statement. What happens if I'm not using a database in my application? It's always checking the `Schema` to see if it exists.

I'd also correctly name the database config to `modules` or something on them lines.

andersonpem:
Watching

garbinmarcelo:
was looking for something related, if this PR is added to the package it would be very good

PR #861, opened 2019-10-25
Fix codestyle of path helpers and extend the module_path helper

RSpeekenbrink:
Extends the module_path helper with an optional $path parameter to be able to parse a relative path together with the module path. I.E.:

```
$myModuleConfig = module_path('MyModule', 'config/config.php');
```

And fix the codestyle of the other path helpers.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -1,11 +1,11 @@
 <?php
 
 if (! function_exists('module_path')) {
-    function module_path($name)
+    function module_path($name, $path = '')
     {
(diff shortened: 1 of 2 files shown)


nWidart:
Thank you, nice addition 👌 

I think there's suppose to be spaces before/after the dots using psr1/2.

RSpeekenbrink:
Updated 👌

nWidart:
Thanks! 👍

PR #900, opened 2020-01-12
Fix view publishing and view cache issue

rabol:
With this PR two issues will be solved

It is possible to publish views using php artisan command and view caching is possible

### DIFF
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -59,11 +59,14 @@ class $CLASS$ extends ServiceProvider
 
         $this->publishes([
             $sourcePath => $viewPath
-        ],'views');
+        ],'modules-$LOWER_NAME$-views');
 
-        $this->loadViewsFrom(array_merge(array_map(function ($path) {
-            return $path . '/modules/$LOWER_NAME$';
-        }, \Config::get('view.paths')), [$sourcePath]), '$LOWER_NAME$');
+        // Load the view from the module path or from the published path
+        if (file_exists($viewPath)) {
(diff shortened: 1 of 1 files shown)


nWidart:
Thanks.

It seems like this might already be working before.
```php
        $this->loadViewsFrom(array_merge(array_map(function ($path) {
            return $path . '/modules/$LOWER_NAME$';
        }, \Config::get('view.paths')), [$sourcePath]), '$LOWER_NAME$');
```

and `view.paths` is an array you can customise. 

```php
    'paths' => [
        resource_path('views'),
    ],
```

rabol:
Hi

no, the original did not work as i would add one or more empty directories.

The original implementation add two:
```
<base_dir>/resources/views/modules/<module_name>
<base_dir>/Modules/<Module_name>/Resources/views
```
the first directory path does not exists so therefore the artisan view:cache commands fail

the 'tag' for publishing was equal for all modules which is why the artisan vendor:publish could not figure out what to publish.

nWidart:
I just tested it out and it does work.

The views need to exist under the correct path as per the SP: `$viewPath = resource_path('views/modules/blog');`

image

The views are also properly published using:
```
vendor:publish --tag=views 
```
image

rabol:
But you are using the same tag for all modules so either the wrong or all views will be published - which might not be what the users want

It’s your package, do as you see fit

nWidart:
Indeed, I've added the extra tag, instead of replacing it, which would be a breaking change potentially.

https://github.com/nWidart/laravel-modules/commit/c5c078313ce218bc15accfa7c0054597b84d6893

Issue #747, opened 2019-04-04
[Feature Request] Add make-observer command


sykorax:
Would be useful to have [this](https://laravel.com/docs/5.8/eloquent#observers) one in the package.

nWidart:
Hello,

Why not indeed. I'd accept a PR on this feature with tests. 😄

nWidart:
I'll close this as there hasn't been a reply in a while, feel free to comment if needed.

ugurdnlr:
Hello,

Has this arrangement been added? Because I need this.

4n70w4:
+1

4n70w4:
https://github.com/nWidart/laravel-modules/pull/915

PR #915, opened 2020-01-29
#747 add module:make-observer

4n70w4:


src/Commands/ObserverMakeCommand.php
```diff
@@ -0,0 +1,171 @@
...
+    private function getNamespacedModel()
+    {
+        $model = $this->option('model');
+
+        $model = str_replace('/', '\\', $model);
+
+        $module = $this->laravel['modules']->findOrFail($this->getModuleName());
+        $namespaceModel = $this->getClassNamespace($module, $this->getModelNamespace() ) . '\\' . $model;
+
+        if (Str::startsWith($model, '\\')) {
```
nWidart: Why is all this needed?

src/Commands/GeneratorCommand.php
```diff
@@ -76,10 +76,11 @@ public function getDefaultNamespace() : string
      * Get class namespace.
      *
      * @param \Nwidart\Modules\Module $module
+     * @param string $concrete
      *
      * @return string
      */
-    public function getClassNamespace($module)
+    public function getClassNamespace($module, string $concrete = null)
```
nWidart: Why is this argument needed? All other commands were able to make it work without it.

src/Commands/stubs/scaffold/provider.stub
```diff
@@ -18,6 +18,7 @@ class $CLASS$ extends ServiceProvider
         $this->registerConfig();
         $this->registerViews();
         $this->registerFactories();
+        $this->registerObservers();
```
nWidart: I would prefer if this was not present by default. Instead, have this documented that it's required to add to have them loaded.

tests/Commands/__snapshots__/ProviderMakeCommandTest__it_can_change_the_default_namespace_specific__1.php
```diff
@@ -18,6 +18,7 @@ public function boot()
         $this->registerConfig();
         $this->registerViews();
         $this->registerFactories();
+		$this->registerObservers();
```
nWidart: Your generated stubs somehow are using tabs instead of spaces I see.

src/Commands/GeneratorCommand.php
4n70w4: Other commands work with one namespace, this command need to generate a few different namespaces.

src/Commands/ObserverMakeCommand.php
4n70w4: Generate namespace for model, of cource. Do you know a better way?

src/Commands/stubs/scaffold/provider.stub
4n70w4: Firstly it is useful. Minimizes manual work.

I can say the same thing about `$this->registerViews();` and `$this->registerFactories();`.
In my case, I need observers, instead of views and factories.

src/Commands/ObserverMakeCommand.php
nWidart: I got that part thanks. I'm asking why replacing slashes.

src/Commands/stubs/scaffold/provider.stub
nWidart: I understand why it's useful. I'm asking because not a lot of people use these, hence why there's not much demand for this feature.

Do you think more people use observers than views?

4n70w4: I only develop a backend. I have only observers. I don't have a view. I just leave empty `registerViews()`, even if I don’t have views. Those who do not have observers can do the same. It costs nothing in performance. It seems like a compromise.

src/Commands/ObserverMakeCommand.php
4n70w4: I will take a look and say later. But without this, it didn’t work.

src/Commands/stubs/scaffold/provider.stub
nWidart: I understand your need, I'm asking if you think that's what most people have.

4n70w4: I think that people who have grown up to dividing the application into modules have also grown up to using observers. People who do not use observers are likely to not divide applications into modules. I began to use observers more actively when I began to divide the application into modules. With observers it is useful to hang the functionality of one module on the model of another module. The division into modules encourages you to make more servers.

nWidart: I don't use observers.

JulienTant: laravel-module exists since 2016, and in 4 years nobody requested anything regarding observers. You're right in the sense that observers are awesome, and maybe everybody should use them, turns out this is not the case so far.

You should consider removing this `$this->registerObservers();` and documenting it. If people come complaning that this is not active by default i'm sure @nWidart will be pleased to turn it on by default.

Khorsan:
Hey, everyone. Is this ready to be merged (except for the conflicts)?

nWidart:
It still has unresolved comments. tabs vs spaces, etc.

AEK-BKF:
Hello,
Any updates here ?  :))

4n70w4:
@AEK-BKF you can use my fork to use this feature.

PR #936, opened 2020-03-08
Make properties of frequently used scaffolding placeholders

bert-w:
The previous stub copied a lot of `'$MODULE$'` and `'$LOWER_NAME$'` into the scaffold, which imho introduces bugs more easily when you have to rename a module after creating a new one. This setup is a little bit cleaner.

I didn't bother to change the other placeholders as properties, since those weren't copied as frequently.

#### Test case
Run `php artisan module:make TestModule` and check the outputs of the generated `TestModuleServiceProvider.php`.

### DIFF
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -8,6 +8,16 @@ use Illuminate\Database\Eloquent\Factory;
 class $CLASS$ extends ServiceProvider
 {
     /**
+     * @var string $moduleName
+     */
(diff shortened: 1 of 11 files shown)


nWidart:
Thank you for the contribution, this is looking good.

Tests are failing because of the snapshots, you can update them by running `./vendor/bin/phpunit --no-coverage -d --update-snapshots`

bert-w:
Alright it should be good now, I havent used snapshots yet, seems like an easy way to run a lot of tests :)

nWidart:
Thanks!

PR #940, opened 2020-03-11
[Feature Request ]Enable or disable all modules

pickmanmurimi:
Currently if i have like 15 modules, i have to enable or disable each one by one.
I have added a command to module:enable-all and modules:disable-all.
What do you think of such a feature?


nWidart:
Hello,

Thank you for the suggestion and quick pull request!
However, I think this might be better suited inside the `EnableCommand`. If for example, we don't provide any module name, it'll run for all modules.
Other commands work similarly.

pickmanmurimi:
Hi,
Thanks for the quick reply.
Nice, I think having the login in the enable command will be great. 
Lemme edit and push.

scancode-system:
So let me understand, in new version removed and enable all module with one command will be available? Will the cade that pcikmanmutimi merge with the core of laravel module packge?

pickmanmurimi:
@scancode-system If you run `module:enable` or `module:disable` with no arguments, it will enable all or disable all modules

scancode-system:
curently show this message "Not enough arguments (missing: "module")."

pickmanmurimi:
@nwidart is yet to review and merge the feature. Maybe he can shed some light into it and when it can be merged.

scancode-system:
For a moment i put a this feature by myself in a module that a made. But its much better to use this feature in native way from packge than my own

pickmanmurimi:
@scancode-system that's a great idea, i'll do it on a module for now, but like you said , it would be nice to have the feature supported on a native level.

nWidart:
Oops sorry I missed you updated the PR. It looks good, I'll apply the code style on my side. 👍

pickmanmurimi:
Great thanks @nWidart

scancode-system:
nWidart Yehh put there please. It will be nice to see that on further versions. One Thing, i was searching way to check if module is enable or disable but i cant see any function to do that at this page

https://nwidart.com/laravel-modules/v6/advanced-tools/module-methods

Is there other way to do that?

pickmanmurimi:
@scancode-system are you looking for `module:list`

PR #973, opened 2020-04-23
cs/phpdoc fix Commands

RomanMin:


nWidart:
Hello,

Thank you but the builds are failing. 
I'm also not a fan of adding these just for the sake of adding them.
I would prefer adding proper return types over these doc blocks.
Also on the command properties, they were removed on purpose.

PR #978, opened 2020-05-03
Update EnableCommand.php

jnbn:
Fixes a bug that causes Nwidart\Modules\FileRepository::findOrFail() must be of the type string, null given.

### DIFF
--- a/src/Commands/EnableCommand.php
+++ b/src/Commands/EnableCommand.php
@@ -31,7 +31,7 @@ class EnableCommand extends Command
          * check if user entred an argument
          */
         if ($this->argument('module') === null) {
-            $this->enableAll();
+            return $this->enableAll();
         }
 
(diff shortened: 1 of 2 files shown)


nWidart:
Hi,

Could you add a test for this case, there's a [EnableCommandTest[(https://github.com/nWidart/laravel-modules/blob/master/tests/Commands/EnableCommandTest.php). 
This would make sure this error doesn't occur in the future.

jnbn:
Created a test case for enabling all modules.
Added and Taxonomy module as secondary module for it.

Getting 4 notifications for assertDirectoryNotExists() as it's deprecated in PHPUnit10. So I'm gonna change them to  assertDirectoryDoesNotExist() on next PR

nWidart:
Awesome! Thanks again 👌

PR #980, opened 2020-05-08
PHPUnit10 Support

jnbn:
renamed assertDirectoryNotExists and assertFileNotExists to assertDirectoryDoesNotExist and assertFileDoesNotExist as they'll be deprecated on PHPUnit 10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "ext-json": "*"
   },
   "require-dev": {
-    "phpunit/phpunit": "^8.5",
+    "phpunit/phpunit": "^8.5 || ^9.0",
     "mockery/mockery": "~1.0",
     "orchestra/testbench": "^5.0",
     "friendsofphp/php-cs-fixer": "^2.16",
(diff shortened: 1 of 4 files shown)


jnbn:
Not sure I we should upgrade phpunit dependancy directly or add it with || to composer.

Tested with 9.1.4

mikemand:
Since it's a dev dependency of this package, it won't have any effect on anyone installing the package. I don't see anything wrong with upgrading PHPUnit here, other than we are still testing older versions of PHP. (I say older... I remember the days of PHP 5.3, so PHP 7.2 really isn't that old in my opinion, LOL.)

PHPUnit 9 only supports PHP 7.3 and 7.4. So for the PHP 7.2 tests, we still need PHPUnit 8.

jnbn:
Oh I see..

I fails as that method doesn't exist on previous versions. 

So maybe we shooul close this PR or merge it after dropping support for older versions.

nWidart:
Hi,

Good idea yeah @jnbn , we'll do that 😄

PR #983, opened 2020-05-12
moved routeServiceProvider Registration to boot method in provider.stub

rahulniraula:
Registering RouteServiceProvider in the boot method gives ability to inject whatever laravel provides.
Currently, RouteServiceProvider is registered before the application boots up and we cant use config() and other functions in the routes.


nWidart:
Tests are failing so this can't be merged until then.

PR #996, opened 2020-05-25
Removing module service provider from composer.json stub

oleaass:
Having the service provider in composer.json causes Laravel to auto detect it. This change stops this from happening


nWidart:
Sorry for the delay, thanks for the contribution! 👌

PR #1011, opened 2020-06-26
Publish module config file without updating the module provider.

narik41:
Publish the module config file without updating the module provider.
Read the module config file using following :
	config("modulename::configfilename.key")


nWidart:
Hi,

Thank you for the contribution! However I'd prefer not to merge this as looping over a folder consistently, will have a performance cost. I think it's better to leave it fully manual and flexible for people.

PR #1016, opened 2020-06-30
Fixed custom stub path issue.

maab16:
Current version of laravel-modules doesn't work custom stub path because of LaravelModulesServiceProvider used fixed path on line number 35.
```
Stub::setBasePath(__DIR__ . '/Commands/stubs');
```
Replace above code with this code then it works fine.
```
$path = $this->app['config']->get('modules.stubs.path') ?? __DIR__ . '/Commands/stubs';
        Stub::setBasePath($path);
```


nWidart:
Thank you! Good catch! 👌

PR #1020, opened 2020-07-05
Fix controller return type

Esirei:
Hello, this PR is a fix for the return type of a normal controller's methods.


nWidart:
When was this changed in laravel?

Esirei:
Since v6 I believe, never got the chance to use that before 7 was released, so might be wrong.

nWidart:
Okay cool :)

Issue #1017, opened 2020-07-03
Make listener CLI with --event flag wrong namespace


Norgul:
Using the following:

    php artisan module:make-listener SendsEmails Email --event=EmailInitialized --queued

produces the right class, however it is referencing:

    use Modules\Email\Listeners\Events\EmailInitialized;

instead of 

    use Modules\Email\Events\EmailInitialized;

nWidart:
Thanks, this has been fixed in the linked PR. 👍

PR #1033, opened 2020-08-01
Refactor ListenerMakeCommand

rcbgalido:
This PR fixes issue #1017 where the Event class namespace should depend on the Event path set in the config file. Also added support for Event classes placed on sub-directories.


nWidart:
Thanks!

PR #1061, opened 2020-09-12
Laravel 8 - Factories Workaround

midhunmonachan:
I found a way to make the module factories work by making change in ServiceProvider class within the module

Changed `app(Factory::class)->load(module_path($this->moduleName, '$FACTORIES_PATH$'));` to `$this->loadFactoriesFrom(module_path($this->moduleName, '$FACTORIES_PATH$'));`

I traced back to the project and this stub file is where I found this code. I've made change to the same file in this pull request. 

Hope it helps and works! 

Cheers

### DIFF
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -98,7 +98,7 @@ class $CLASS$ extends ServiceProvider
     public function registerFactories()
     {
         if (! app()->environment('production') && $this->app->runningInConsole()) {
-            app(Factory::class)->load(module_path($this->moduleName, '$FACTORIES_PATH$'));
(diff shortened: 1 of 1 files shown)


vijitail:
`loadFactoriesFrom()` is deprecated. I guess for now installing `laravel/legacy-factories` would do the trick.

Norgul:
If you're using old factories then install `legacy-factories`. If you're using new factories then those are namespaced, so you should just comment out `$this->registerFactories();` in each module service provider

nWidart:
Thank you for the PR, I've decided to remove autoloading factories by default in the laravel 8 support PR https://github.com/nWidart/laravel-modules/pull/1073

PR #1085, opened 2020-10-06
Command management

lindo-llc:
Each command can be enabled/disabled from within the configuration file, allowing for more flexibility


nWidart:
I like it, very nice!

nWidart:
Tests are failing so make sure to run them locally to see why. (I m guessing maybe commands not considered enabled in tests env, you can set that up with orchestra testbanch

lindo-llc:
> Tests are failing so make sure to run them locally to see why. (I m guessing maybe commands not considered enabled in tests env, you can set that up with orchestra testbench

Not sure who or what have been doing, but this is the error:
`182) Nwidart\Modules\Tests\LaravelModulesServiceProviderTest::it_throws_exception_if_config_is_invalid
Error: Class 'Modules\WrongModule\Providers\WrongModuleServiceProvider' not found`

nWidart:
Ah that's when a test stopped in the middle, you need to remove the module from the testbench folder:
`vendor/orchestra/testbench-core/laravel/modules`

nWidart:
In the `BaseTestCase` you can add configuration for the test environment.

lindo-llc:
Man, I'm slowly loosing patience with this `WrongModule` 😂

nWidart:
Haha yeah, it's annoying it doesn't "cleanup" after a crash during a test.

lindo-llc:
Alright, got rid of it, but now some weird testing errors popping up, like symlink() permission denied, mkdir() file exists...

nWidart:
hm strange, make sure that laravel/modules folder is empty before running the tests.

lindo-llc:
Should pass now

nWidart:
I see the tests seem to be passing now 👌 
The `::class` notation didn't work?

lindo-llc:
I was trying to figure out the testing errors. I will take a look into that now

src/Providers/ConsoleServiceProvider.php
```diff
@@ -101,7 +110,34 @@ class ConsoleServiceProvider extends ServiceProvider
...
+        return $commands;
+    }
+
+    /**
+     * Resolve command namespace
+     *
+     * @param string $command
+     * @return void
+     */
+    private function resolveCommandNamespace(string $command)
```
nWidart, line 138: Is this still needed when using `::class` ?

lindo-llc: Because `ConsoleServiceProvider` NEEDS to use `USE` for the classes for conflicting namespaces this check is needed.

Example: `Nwidart\Modules\Providers\Nwidart\Modules\Console\CommandName`

This lets the config to be cleaner.

Although, I will merge it into one method.

lindo-llc: Because it uses `::class` notation it conflicts

nWidart:
Awesome, thanks!

PR #1088, opened 2020-10-07
Non-merger: WIP: Extended Commands

lindo-llc:


FoksVHox:
Maybe acutallly provide a description?

lindo-llc:
Still working on this. Seeing if these shitty tests would pass, obviously not

FoksVHox:
> Still working on this. Seeing if these shitty tests would pass, obviously not

Mark it as a wip, I believe that the test things should work as it does now when you've marked it as a wip

PR #1089, opened 2020-10-07
Command signature patch

msztorc:
- Updated command stub (changed property $name to $signature)
- Updated test snapshots

### DIFF
--- a/src/Commands/stubs/command.stub
+++ b/src/Commands/stubs/command.stub
@@ -9,11 +9,11 @@ use Symfony\Component\Console\Input\InputArgument;
 class $CLASS$ extends Command
 {
     /**
-     * The console command name.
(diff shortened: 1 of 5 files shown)


nWidart:
Great, thanks!
Thanks for fixing the tests too 👍

lindo-llc:
Why was this needed?

msztorc:
Property `$signature` is more appropriate than `$name` because except defining the command name, it allows you also to use named options like: `{--queue=}`

Additionally, `$signature` property is added by default in Laravel commands, see here: https://laravel.com/docs/8.x/artisan

nWidart:
I prefer the "old" way of writing the signatures, but I agree this is the default now so.. 🤷‍♂️

Issue #1086, opened 2020-10-06
Add a module:make-component


FoksVHox:
Laravel has a component command, which is really awesome, and I think that people would like to see the same functionality in this package. 

My use case is that I'm making a Laravel vapor-like site, and I'd like to use the newly added dynamic-component in one of my modules.

magic-thomas:
+1

FoksVHox:
#1110 Fixes this issue

FoksVHox:
Closed in 850931b

PR #1090, opened 2020-10-08
Add command for creating a component view

FoksVHox:
*Needs some more work, as it only creates the view and not the class...* 

Solves #1086

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -126,6 +126,7 @@ return [
             'emails' => ['path' => 'Emails', 'generate' => false],
             'notifications' => ['path' => 'Notifications', 'generate' => false],
             'resource' => ['path' => 'Transformers', 'generate' => false],
+            'view-component' => ['path' => 'Resources/views/components', 'generate' => false],
         ],
     ],
 
--- a/src/Commands/ComponentMakeCommand.php
+++ b/src/Commands/ComponentMakeCommand.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Nwidart\Modules\Commands;
+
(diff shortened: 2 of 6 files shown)


FoksVHox:
@nWidart I'm not sure what you think about changing the `GeneratorCommand`, so it looks more like the one form Laravel, the reason being that I need to create a component class for the views.

nWidart:
hi @FoksVHox , what exactly would change? What would it allow that it can't do now?

Thanks!

FoksVHox:
> hi @FoksVHox , what exactly would change? What would it allow that it can't do now?
> 
> Thanks!

Right now I can't make the Component Class which is used for them in Laravel. I want to change it, so I'm able to create the Component Class.

nWidart:
I understand, but what is it that blocks you from creating the component class, specifically?

FoksVHox:
> I understand, but what is it that blocks y

As far as I'm aware of I'm not able to create more files inside the `getTemplateContents` function.

nWidart:
Ah ok I see. You could have multiple sub commands that are being called by the main generate one for this PR, but that might be too hacky. So if you have a good solution by using laravel's structure, while keeping all existing commands working, why not.

FoksVHox:
> Ah ok I see. You could have multiple sub commands that are being called by the main generate one for this PR, but that might be too hacky. So if you have a good solution by using laravel's structure, while keeping all existing commands working, why not.
I'll play around with it

FoksVHox:
Closing in favor of #1110

PR #1091, opened 2020-10-09
Updated list command to use getPriority method and priority property

semsphy:


nWidart:
Ooh thanks, good catch! 👌

PR #1100, opened 2020-10-13
added Laravel 8 compatible Factory and Model generation

Rezrazi:
changed the `module:make-factory` argument description from `ModelFactory` to `Model` that way the command generates a factory with a model name rather that a factory name as argument

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ build
 coverage
 .php_cs.cache
 .phpunit.result.cache
+
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,3 @@ build
 coverage
 .php_cs.cache
 .phpunit.result.cache
-
-.idea
--- a/src/Commands/stubs/model.stub
+++ b/src/Commands/stubs/model.stub
@@ -3,8 +3,16 @@
 namespace $NAMESPACE$;
 
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ build
 coverage
 .php_cs.cache
 .phpunit.result.cache
+
+.idea
```
nWidart, line 10: Can you remove this please?

src/Commands/stubs/model.stub
```diff
@@ -3,8 +3,16 @@
 namespace $NAMESPACE$;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
 
 class $CLASS$ extends Model
 {
+    use HasFactory;
+
+    protected static function newFactory()
```
nWidart, line 12: I think this should probably below the fillable properties, to keep properties on top.

nWidart (commented):
Thanks for the PR!

Rezrazi (commented):
applied the following changes :
- updated `.gitignore` to remove IDE specific lines
- updated `model.stub` to fix the method order

Rezrazi (commented):
updated the snapshots to get the `ModelMakeCommandTest` tests to pass

nWidart:
Looks great thanks for the changes!

PR #1102, opened 2020-10-18
compatible with lumen framework

ivi3:


nWidart:
Hi,

Thank you for the pull request, however, I would prefer not to add this as last time someone added lumen support, it caused more issues than it solved. As I had to maintain it, and I myself don't use lumen.

ivi3:
Hi,

ivi3:
Hi,
But i needed this feature on my project and with this changes worked for me without any problem..you can see ivi3/laravel-modules on packagist for check it on lumen framework.
if you think this feature needed some Unit test file and or anything dependencies for accept this feature to project, i can work on. however thank for this great package.

PR #1106, opened 2020-10-24
feat: Silent composer

lindo-llc:
Ability to disable composer output via `config.php` when using `php artisan module:update`, `php artisan module:install`.
`config.php`: `'composer-output' => false`


nWidart:
Thanks!

Issue #947, opened 2020-03-20
Define Controller type when creating the module


garbinmarcelo:
Hello,
The package offers 3 types of controllers, web, api and plain.

Would it be possible when a new module is created to define the type of controller to be created? (I checked the documentation and has only the --plain)

Example, I use the package for creating api. In this case there would be a command like: 
`php artisan module:make User --api`

I believe that this would be a useful resource, you could consider that for v7... :smile:

kkornwinkel:
Hi,
I agree with @garbinmarcelo
That would be a great feature for model-based automated API creation

nWidart:
Hello,

Sorry for the late reply. I like this idea as well.

The default could stay as is, and then

- module:make User --web
- module:make User --api
- module:make User --plain

for example

PR #1109, opened 2020-10-28
Feature/define controller type

ahmedggalal:
add the ability to create modules using :
module:make User --web
module:make User --api
module:make User --plain

### DIFF
--- a/src/Commands/ModuleMakeCommand.php
+++ b/src/Commands/ModuleMakeCommand.php
@@ -33,7 +33,6 @@ class ModuleMakeCommand extends Command
         $success = true;
 
         foreach ($names as $name) {
-            $type = $this->getModuleType($this->option('plain'),$this->option('api'));
             $code = with(new ModuleGenerator($name))
                 ->setFilesystem($this->laravel['files'])
                 ->setModule($this->laravel['modules'])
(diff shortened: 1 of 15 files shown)


src/Commands/ModuleMakeCommand.php
```diff
@@ -68,8 +69,14 @@ protected function getOptions()
...
             ['plain', 'p', InputOption::VALUE_NONE, 'Generate a plain module (without some resources).'],
+            ['api', null, InputOption::VALUE_NONE, 'Generate an api module.'],
+            ['web', null, InputOption::VALUE_NONE, 'Generate a web module.'],
             ['disabled', 'd', InputOption::VALUE_NONE, 'Do not enable the module at creation.'],
             ['force', null, InputOption::VALUE_NONE, 'Force the operation to run when the module already exists.'],
         ];
     }
+
+    private function getModuleType($isPlain,$isApi,$isWeb){
+        return ($isPlain && $isApi && $isWeb)||($isPlain && $isApi)||($isApi && $isWeb)||($isPlain && $isWeb)?'web':($isPlain&&(!$isApi&&!$isWeb)?'plain':($isApi&&(!$isPlain&&!$isWeb)?'api':'web'));
```
nWidart, line 80: I have no clue wtf this does.

ahmedggalal: it just returns web if users entered more that one options , --plain --web --api

nWidart: yes I guessed that, but it's unreadable. The new version is still unreadable. Don't use ternaries here.
Also please create tests for each possible outcome.

nWidart:
Thank you for the changes! 👍

ahmedggalal:
#947

PR #1119, opened 2020-11-10
Fix getting the full class name

mtawil:


nWidart:
Hello,

What issue does this fix?
I'm able to use module commands without it.

mtawil:
The issue that there is a missing class name, you're calling `Nwidart\Modules\Commands\`, just namespace only without a class name.

When I clear eveything using:
```
rm -f core/bootstrap/cache/*                                                         
php artisan clear-compiled
php artisan view:cache
php artisan optimize
php artisan config:clear
```

I will get this error when trying to do something like `composer dump-autoload`:
image

On my development environment, this change is fixing the issue, but on the production, it fixed by downgrading to v8.0.0.

nWidart:
I see okay, thanks!

PR #1128, opened 2020-11-16
Fixed the ability to override commands via config file.

mikemand:
The way the `resolveCommands()` method was working, if I were to replace one of the commands in the config with my own it wouldn't work. Since the method was prepending `Nwidart\Modules\Commands` to the class if it didn't already exist, I would have to make my commands in that namespace which would break composer autoloading.

**This is a breaking change.** If someone doesn't have the corrected import and class-based entries in their config file for the list of commands, it will crash while looking for commands without their correct namespace. If someone doesn't have the `commands` array in their config already, it would not change anything for them.


nWidart:
Oh wow yes, I missed that. I'm wondering, why don't we use the full namespace in the config file?

mikemand:
We can, if you want too. It's shorter this way. PHP allows you to `use` an entire namespace (in this case, by doing `use Nwidart\Modules\Commands;` I am telling PHP that I want to use any class in that namespace), and then you only have to reference part of that namespace when using a class. Basically, it's just a shortcut so we don't have to `use` every single command at the top of the file; which gets long and unwieldy.

nWidart:
ah yea ok, was just making sure that users can now fully overwrite commands (ie the class isn't hardcoded on the loop)

mikemand:
How do you want to handle the breaking change? A new major version (v9), or something else?

PR #1137, opened 2020-12-07
fix: feat: Silent composer

lindo-llc:
Fixes an incorrectly placed configuration key `composer-output`


nWidart:
Good catch, thanks!

PR #1105, opened 2020-10-23
Configuration to change the default stub file variable style

lindo-llc:
Ability to change stub file variable style. This provides the ability to change the default style of `$` (dollar sign) to anything. 
Examples include: `{{ VARIABLE }}`, `[[VARIABLE]]`, or anything you can think of, even  `~~|~VARIABLE~|~~`

PR #1138, opened 2020-12-07
Feat: custom stub variable

lindo-llc:
Another take on:  [#1105 ](https://github.com/nWidart/laravel-modules/pull/1105)

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -27,6 +27,7 @@ return [
 
     'stubs' => [
         'enabled' => false,
+        'style' => '$VARIABLE_NAME$',
         'path' => base_path('vendor/nwidart/laravel-modules/src/Commands/stubs'),
         'files' => [
             'routes/web' => 'Routes/web.php',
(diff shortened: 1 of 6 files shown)


nWidart:
Nice, the tests are green! 😅 
I'm curious, what's the use-case of this again? I understand it was to allow custom 'variable styles', but why exactly?

lindo-llc:
As it stands, the package isn't too flexible. Therefore if implementing some sort of in-app logic that uses this package you have a different style of stubs and are writing a custom logic that uses the package Generator, the generator cannot replace the variables inside the custom stub files, not until you change them in a specific way, that this package requires

Snippet example of a custom factory stub file: 
`<?php

namespace {{ NAMESPACE }};
{{ MODEL_NAMESPACE }}
use Illuminate\Database\Eloquent\Factories\Factory;

class {{ FACTORY_NAME }} extends Factory
{
    /**
     * The name of the factory's corresponding model.
     *
     * @var string
     */
    protected $model = {{ MODEL_NAME }}::class;

    ...`

PR #1148, opened 2020-12-17
Fix missing commands

FoksVHox:
This PR fixes the issue with these commands not getting registered:
`module:make-component <name> <module>` 
`module:make-component-view <name> <module>`


nWidart:
Awesome thanks!

PR #1153, opened 2020-12-21
Only support supported Laravel versions

FoksVHox:


mauriciovillam:
I see why, but why?

FoksVHox:
> I see why, but why?

Cause it doesn't make sense to support unsupported laravel versions, that's dumb

mauriciovillam:
Sorry, I may be seeing things wrongly but I don't see in which way a compatible versions cheatsheet is dumb or why this pull request is necessary.

FoksVHox:
There are nobody who used Laravel 5.4 today, as Laravel 5.4 is not supported. I think you should take a look into which versions of Laravel, the Laravel Team support. All versions below Laravel 6 has reached end-of-life.

FoksVHox:
@nWidart What about this, shouldn't we have merged this as well?

nWidart:
Not a fan of this one. The versions array there isn't a list of supported versions, just a matrix of package version -> laravel version.
In case someone is still using laravel 5.6, they know what to use 😅

FoksVHox:
> Not a fan of this one. The versions array there isn't a list of supported versions, just a matrix of package version -> laravel version.
> In case someone is still using laravel 5.6, they know what to use 😅

Oh okay.

Issue #1054, opened 2020-09-11
ERROR: File does not exist at path /var/app/staging/app/Modules/**/module.json


jpswade:
Like [this issue](https://github.com/nWidart/laravel-modules/issues/233), we're also seeing performance issues.

We've turned on cache but, the way our deployment works, is it goes into /var/app/staging/* before moving to /var/app/current/*.

This perfect storm results in this error:

```
[2020-09-11 15:32:12] testing.ERROR: File does not exist at path /var/app/staging/app/Modules/**/module.json. {"exception":"[object] (Illuminate\\Contracts\\Filesystem\\FileNotFoundException(code: 0): File does not exist at path /var/app/staging/app/Modules/**/module.json. at /var/app/current/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:54)
...
```

My view is that we need a post deployment command to reset the cache, but, there isn't one.

Suggestions welcome.

jpswade:
One way would be to add a enable/disable in migration.

nWidart:
Hi,

Not sure about your setup. You might have to clear the cache indeed on every deploy (good practice anyway), maybe symlink something too?

jpswade:
@nWidart I'm still seeing this issue.

The way elasticbeanstalk works is a pretty common "zero downtime" approach, however it does differ to the way envoyer works.

Envoyer has a directory called current, then symlinks to the current release directory.

ElasticBeanstalk on the other hand will create a "staging" directory, then rename it to "current" when it's ready. This behaviour cannot be changed.

For whatever reason this package appears to be caching the full path to the modules.json, rather than a relative one so when the switch happens, it's reading from the wrong directory.

As part of the deployment process, you have to run artisan commands before the switch is done (eg: php artisan migrate), which is when it's still in "staging" so I think this is when it's caching the path.

Post deploy it will then run artisan commands again.

We've had to write a command that will force the clearing of the cache so that the path isn't cached. However there appears to be a bit of a delay so sometimes processes may hit the application before the system has cleared the modules cache.

Caching seems pretty essential to the performance of the library, yet seems to be supported in a very limited way (ie: no native way to clear it automatically) and fixed path rather than relative, none of which is useful when you are deploying quite often.

Is this something you intend to improve upon or has this package reached maturity?

jpswade:
One idea I'm going to try is to see if the config can be updated to be relative...

[...]

PR #1168, opened 2021-01-19
Bugfix/str 4114 illuminatecontractsfilesystemfi

jpswade:
This offers four methods to solve the issued outlined here:

https://github.com/nWidart/laravel-modules/issues/1054

1. We now have a command that we can run to clear the cache (by force, overriding what's in the config) if we need to.
2. If we don't have a modules.json file, then we're probably in the wrong path, so clearing the cache will fix this. Failing that, we (more) gracefully throw an exception, which can be caught, rather than a fatal error we had previously.
3. You can pass in a MODULES_PATH now as an environment variable and specify that if you need to via a .env file
4. It will now handle both paths that are full paths and relative paths, either will work

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -230,7 +230,7 @@ return [
     |
     */
     'cache' => [
-        'enabled' => false,
+        'enabled' => env('MODULES_CACHE_ENABLED', false),
         'key' => 'laravel-modules',
         'lifetime' => 60,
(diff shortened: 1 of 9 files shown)


lamtranb:
Updating module.json as a const is good.
But can we just clear cache in application level ?

Issue #1007, opened 2020-06-20
lumen issue


fayroz:
Hi,
how can i fix this issue with lumen 7 
Class 'Illuminate\Foundation\Support\Providers\RouteServiceProvider' not found

nWidart:
Hi,

Sorry, Lumen is not officially supported. 
You'll have to make it work on your own, or if somebody else can point you in the right direction.

fayroz:
@nWidart  thanks

arturmamedov:
> @nWidart  Are you mean that we not can create modules in lumen? 

You can if you can deal with this by your self, but the package laravel-modules don't do this officially

Issue #955, opened 2020-04-04
Class RouteServiceProvider not found


ognistyi:
Composer:
`"laravel/lumen-framework": "^6.0",
"nwidart/laravel-modules": "6.2.0",`

`> php artisan module:make TestModule`

Open in browser:
Class 'Illuminate\Foundation\Support\Providers\RouteServiceProvider' not found

How to setup on lumen 6.x ?

nWidart:
You forgot to autoload your modules as mentioned in the readme and documentation.

ognistyi:
> You forgot to autoload your modules as mentioned in the readme and documentation.
No, autoload is OK. Lumen 6+ doesn't has illuminate/foundation/*. But it used in module template.

nWidart:
Ah ok. I'm not surehow lumenn does autoloading, this package isn't fully supported in Lumen. Support was added by a community member, but as I don't and never used lumen in the past I can't support that feature.

BehroozBvk:
> Ah ok. I'm not surehow lumenn does autoloading, this package isn't fully supported in Lumen. Support was added by a community member, but as I don't and never used lumen in the past I can't support that feature.

I made the changes and the problem was apparently fixed. I think make the changes in the repository.
If there is a problem, we will solve it with the support of other programmers.

Thanks

dharmendrasha:
Still facing this issue

SubjectXXX:
I see 

> > Ah ok. I'm not surehow lumenn does autoloading, this package isn't fully supported in Lumen. Support was added by a community member, but as I don't and never used lumen in the past I can't support that feature.
> 
> I made the changes and the problem was apparently fixed. I think make the changes in the repository. If there is a problem, we will solve it with the support of other programmers.
> 
> Thanks @nWidart @Hodel1

I'm still facing same issue, even though I have implemented those 5 changes (3 new files and 2 files changed). Do I need to do anything else or is there a problem with new functionality because you were probably doing it on v6, now v8 is out?

Maybe [Hodel1](https://github.com/Hodel1) would know?

Issue #671, opened 2018-11-28
Using route error in lumen 5.7


sniperwzq:
Hi @nWidart,
I use your project in lumen,when i make new module, lumen throw :Class 'Illuminate\Foundation\Support\Providers\RouteServiceProvider' not found,i say your project,RouteServiceProvider use laravel route,but lumen do not support it, so how can i use it in lumen?
My English is not good.Sorry.
I hope you can help me solve it.
Thank you.

nWidart:
Hello,

Lumen support is currently limited, it's quite possible it's not functional anymore.

Did you check the page about usage in Lumen : https://nwidart.com/laravel-modules/v4/lumen ?

sniperwzq:
yes, i check the page about usage in Lumen : https://nwidart.com/laravel-modules/v4/lumen, i see your project ,you use RouteServiceProvider,but lumen do not support it.Do you plan to support lumen? or Which version supports lumen?
I'm looking forward to your support for Lumen.

nWidart:
I don't use Lumen myself, so I cannot support it. Lumen support was added via a contribution. 
Since then lumen has has issues like possibly the one you're facing now.

I'm considering removing lumen support entirely, unless someone shows up who can fix and support it.

nWidart:
I'll close this as there hasn't been a reply in a while, feel free to comment if needed.

PR #1183, opened 2021-02-10
Fixing Lumen Support for Issues like #671, #955, #1007 .

Hodel1:
Lumen has no RouteServiceProvider and handle that routes not equals to laravel, so i created *.lumen.stub files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
   "scripts": {
     "update-snapshots": "./vendor/bin/phpunit --no-coverage -d --update-snapshots",
     "test": "vendor/bin/phpunit",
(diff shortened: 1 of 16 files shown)


Hodel1:
Hmm i wondering about that unit tests failed after change \Config::get() to config().
In Lumen it seems, i don't have an \Config class.

Hodel1:
Now i had a bit time to setup an fully environment with phpunit to fix that errors. 
Have fun :) 

The only thing what is not working in Lumen is: Components and some things like Mail only after installing few packages in lumen.

nWidart:
Hi,

Thanks but Lumen is not supported anymore. It causes too many possible compatibility issues.

Issue #1184, opened 2021-02-12
Create Model - relation to non existing factory?


Hodel1:
Hi,
when i module:make-model then it generates an Model to an relation to an non existing factory (factory in this case lowercase - is an mistake?).

Maybe it should be possible to generate that relation to an factoy, by make-factory?

Greets

PR #1185, opened 2021-02-16
Make-Model with Factory option - See #1184

Hodel1:
php artisan module:make-model ModelName ModuleName (-m -f) 
-f is meaning with factory relation and make-factory Model 
Without -f is that model without factory relation.


Hodel1:
phpunit is now working

PR #1186, opened 2021-02-16
Feature/submodule support

Hodel1:
Creating Submodules like: Admin/Auth, Admin/User, Core/Configuration, etc.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -200,7 +200,8 @@ return [
     'scan' => [
         'enabled' => false,
         'paths' => [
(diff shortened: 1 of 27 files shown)


Hodel1:
now i fixed that php unit errors :)

dcorazolla:
Good idea... Exactly what i need!

mshamaseen:
Why hasn't this merged? Any problem with it? I was about to do a PR for this feature.

mshamaseen:
I've just tried this, this DOESN'T work.

Issue #1188, opened 2021-02-17
 Command "module:make-component" is not defined.  


quentingosset:
When I try to create a component with the command "php artisan module:make-component Test" I get the error that the command is not defined. Is this normal?

tamasori:
No, but yes. Its a bug in the package, if you publish the config file, you should find a part where all the command classes are listed. The two component related classes are missing. The commands keys value should look something like this:

'commands' => [
        CommandMakeCommand::class,
        ControllerMakeCommand::class,
        DisableCommand::class,
        DumpCommand::class,
        EnableCommand::class,
        EventMakeCommand::class,
        JobMakeCommand::class,
        ListenerMakeCommand::class,
        MailMakeCommand::class,
        MiddlewareMakeCommand::class,
        NotificationMakeCommand::class,
        ProviderMakeCommand::class,
        RouteProviderMakeCommand::class,
        InstallCommand::class,
        ListCommand::class,
        ComponentClassMakeCommand::class,
        ComponentViewMakeCommand::class,
        ModuleDeleteCommand::class,
        ModuleMakeCommand::class,
        FactoryMakeCommand::class,
        PolicyMakeCommand::class,
        RequestMakeCommand::class,
        RuleMakeCommand::class,
        MigrateCommand::class,
        MigrateRefreshCommand::class,
        MigrateResetCommand::class,
        MigrateRollbackCommand::class,
        MigrateStatusCommand::class,
        MigrationMakeCommand::class,
        ModelMakeCommand::class,
        PublishCommand::class,
        PublishConfigurationCommand::class,
        PublishMigrationCommand::class,
        PublishTranslationCommand::class,
        SeedCommand::class,
        SeedMakeCommand::class,
        SetupCommand::class,
        UnUseCommand::class,
        UpdateCommand::class,
        UseCommand::class,
        ResourceMakeCommand::class,
        TestMakeCommand::class,
        LaravelModulesV6Migrator::class,
    ],
    
    The two missing classes are:
        ComponentClassMakeCommand::class
        ComponentViewMakeCommand::class

quentingosset:
However, I have the last updated package with the 2 commands added in the ConsoleServiceProvider :/ 

![image](https://user-images.githubusercontent.com/6824137/108403554-237ed780-721f-11eb-8590-83df47c73ad9.png)

vishalscrimatec:
I am facing the same issue. but on windows  Command "make:component" is not defined.

  Did you mean one of these?
[...]

PR #1190, opened 2021-02-18
Command "module:make-component" is not defined

quentingosset:
The two component creation commands have been forgotten in the modules file. 
#1188


nWidart:
Thanks!

PR #1217, opened 2021-04-21
fix class name when special characters are used the rule name

MohammadMehrabani:
**before:** 
```
php artisan module:make-rule phone-number Test

namespace Modules\Test\Rules;

use Illuminate\Contracts\Validation\Rule;

class phone-number implements Rule
{
```

**after:**
```
php artisan module:make-rule phone-number Test

namespace Modules\Test\Rules;

use Illuminate\Contracts\Validation\Rule;

class PhoneNumber implements Rule
{
```


nWidart:
Thanks!

PR #1241, opened 2021-06-05
Update create command stub

pickmanmurimi:
`getArguments()` and `getOptions()` can only be used when we are not using fluent definition of the `$signature`.

From `Illuminate\Console\Command;` we can only use `getArguments()` and `getOptions()` if we use `$name` to define our command name, else laravel will use fluent definition parsing to figure out your arguments and options,

    From Illuminate\Console\Command
    if (! isset($this->signature)) {
         $this->specifyParameters();
    }
    
 I changed the command stub to use `$name` as the default command name definition. This will allow use of the `getOptions` and `getArguments` functions.
 
 If we still want to use `$signature` to match the default laravel stubs then get rid of the `getOptions` and `getArguments` methods to prevent the confusion this causes.


nWidart:

PR #1242, opened 2021-06-10
Add routes_with_namespace to config file

gorczakw:
PHP callable syntax was added to Laravel 8. Now namespace property in the RouteServiceProvider is nullable by default. When We create a new module, our new RouteServiceProvider file has namespace property with path. This config allows you to create a new module with the namespace property set as null.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -272,4 +272,6 @@ return [
     ],
 
     'activator' => 'file',
+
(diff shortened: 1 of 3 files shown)


nWidart:
Hi,

Thank you for contribution.
I think it's easier for the user to customise the template and remove this line than adding one more configuration option.

PR #1254, opened 2021-07-28
add optional controller flag for making model command

debuqer:
Use controller flag to create and assign controller to a model

`php artisan module:make-model User -mc`

result 

> Created : .../Modules/ModuleName/Entities/User.php
> Created : .../Modules/ModuleName/Database/Migrations/2021_07_28_140048_create_users_table.php
> Created : .../Modules/ModuleName/Http/Controllers/UserController.php
>


nWidart:
Thankns!

PR #1286, opened 2021-09-23
Laravel mix v6 support

ldavidsp:
https://github.com/nWidart/laravel-modules/issues/1285


nWidart:
Hi and thank you for your contribution!

I'm guessing some tests are going to fail, you might have to update those that check the contents of this file.

nWidart:
Great, thanks!

PR #1301, opened 2021-10-15
Add event name prefix by config

BARTlab:
When generating automatic events for modules, a hardcoded prefix is substituted for the name "modules.". This request adds the custom prefix feature


BARTlab:
What is required for this pull request to be accepted?

PR #1313, opened 2021-11-03
Fix error on anonymous migration class

fadlee:
Laravel team introduce anonymous migration in Laravel 8.37, which solves migration class name collision issue.

Currently, the `module:migrate` command runs normally with anonymous migration.

But, `migrate-rollback` & `migrate-reset` trigger `Class not found` error.

So, I propose this changes to fix that issue.

### DIFF
--- a/src/Migrations/Migrator.php
+++ b/src/Migrations/Migrator.php
@@ -198,9 +198,13 @@ class Migrator
      */
     public function resolve($file)
     {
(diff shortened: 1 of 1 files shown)


romfis91:
@nWidart could you please check and approve this PR? Currently it isn't possible to use anonymous migration.

nWidart:
Thanks!

PR #1348, opened 2022-01-27
fix: artisan module:migrate for lumen

leonardohipolito:
fix error 
Nwidart\Modules\Migrations\Migrator::__construct(): Argument #2 ($application) must be of type Illuminate\Contracts\Foundation\Application, Laravel\Lumen\Application given, called in vendor/nwidart/laravel-modules/src/Commands/MigrateCommand.php on line 67


dcblogdev:
this PR fails the tests, can you update with a passing test.

PR #1363, opened 2022-02-22
Anonymous migration class

iamine:
Add Anonymous migration class as default like in Laravel 9.0


dcblogdev:
thanks, can you update the stubs to close the class Anonymous migration classes now should have this signature:

```php
return new class extends Migration 
{
    // ...
};
```

PR #1380, opened 2022-03-28
Adding suffix in module name (Needed when Publishing Modules)

feeh27:
This fix adapts the initial structure to be compatible with the module publishing.

All configurations were made based on the necessary adjustments, according to the documentation at the link below:
https://nwidart.com/laravel-modules/v6/advanced-tools/publishing-modules


feeh27:
I've been working with magento for many years and I know that it somehow automatically loads the namespace of its modules.

After studying a little about composer and how magento uses it, I found out how it is done, the "Added module dynamic autoloader" commit includes this functionality and it is no longer necessary to add it to Laravel's main composer.json

feeh27:
I've been working with magento for a few years now and I know that it somehow automatically loads the namespace of its modules.

After studying a little about composer and how magento uses it, I found out how it was made.

The "Added module dynamic autoloader" commit includes this functionality and it is no longer necessary to add it to the main Laravel composer.json as it will be automatically loaded now.

feeh27:
@dcblogdev and @nWidart, can you help me with this PR?

feeh27:
@dcblogdev ? @nWidart ?

Issue #729, opened 2019-03-11
Is it possible to do nested modules?


benlonsdale:
I can't see any documentation on this, but is there a way to install sub-modules inside a module?

nWidart:
Hi,

Not using any of the modules commands no.
You could do it manually if you want to, but I wouldn't recommend having submodules in the first place.

benlonsdale:
Okay, follow up question then. Are there any examples of including an external module installed through composer?

nWidart:
Yeah, [AsgardCMS](https://github.com/AsgardCms/Platform) does this. Though that's a bit more special since it's using composer plugin (which installs the module in the Modules folder instead of vendor). But you should get the idea from that.

benlonsdale:
Thanks for your help 👍

Issue #578, opened 2018-07-04
Sub-Modules


imagina:
We are creating an small e-commerce module for AsgardCms. Right now we are implementing the shipping, payment, modules, etc. like laravel-modules. But... all of this modules feels like would be better like sub-modules of our main ecommerce module.

What do you think @nWidart ? Any idea for this implementation.

Regards.

nWidart:
Hello,

That thought crossed my mind before too 😄.
I personally don't think having submodules is a good idea, architecture-wise.

I made an e-commerce with asgard recently too and considered your idea at the time.
However, I preferred keeping all modules "one level" only, which makes it easier to re-use parts and not the whole.
For instance, I had a cart module, an order module, a product module, an invoice module, etc.
On another client, I could re-use some modules, but not others. Since all modules were fairly well decoupled it was easy to re-use the ones that matched the need for the second client.

nWidart:
I'll close this as there hasn't been a reply in a while, feel free to comment if needed.

Issue #201, opened 2017-04-18
Sub folders and modules


marcelogarbin:
Hello again,

Is it possible to organize modules within sub folders?

Ex .: Stock/Products

I ran the command: "php artisan module:make Stock/Products"

All folders have been created, but the controller is not created and an error occurs after php artisan is executed.

Error:
  [Nwidart \ Modules \ Exceptions \ ModuleNotFoundException]
   Module [Stock/Products] does not exist!

I do not know if it would be possible and would be correct to create subfolders to better organize the application.

Grateful

nWidart:
Hello,

This is not possible. However, you can add a dependency to your Product module. You can do that using the usual composer way of handling dependencies, via the composer.json file.

marcelogarbin:
Thanks for the answer. 

As a beginner, could you give a practical example (code) about what you cited?

nWidart:
if you look at your module, you have a `composer.json` file inside there.
In the `require` key of that file, you can add you module. Prior to this your module will have to be registered on packagist (or private equivalent like toranproxy/satis/..).

There is also a command `php artisan module:update Product` which will add those dependencies of the product module, onto your main composer.json file, the one on the root of your project.

naco99:
Hello,
I run into this issue for a while, so I just set the "scan.enabled" property to true (in modules.php  config file), and added paths where submodules lives.
But when I need to generate a new subModule, I first change the "modules" property to the parent path, and also the "namespace".
and everything worked fine

inovek:
I think this wouldn't be such a bad feature. I just tried the same and hoped to archive some structure like:

```
Modules/Endpoint/Products
Modules/Endpoint/Orders
Modules/GUI/Products
Modules/GUI/Orders 
```

PR #1409, opened 2022-07-07
Feature/allow submodule

mshamaseen:
Add the ability to create submodules like Tests/Test.

1. I've distinguished between the `name` and `base_name` in src/Generators/ModuleGenerator.php; for instance, for Blog/Sub, the name will be Blog/Sub, but the base name will be Sub.
2. I've considered any directory with a `module.json` a module, so, even if you have a directory that has just `module.json` file, and it is inside `Modeuls` path, thus this is considered a module, farther more, even u have more `module.json` in recursive directory inside `Modules` every one of them will be considered a module.
3. I've created a test for the submodule case.


mshamaseen:
@dcblogdev I'd appreciate your input here.

dcblogdev:
> @dcblogdev I'd appreciate your input here.

I'll review this as soon as I can, I like the idea of this concept, I normally separate modules into top-level directories, and this opens that possibility up more.

mshamaseen:
@dcblogdev I'll fix that soon, thanks for the update.

mshamaseen:
@dcblogdev done, please rerun the tests.

mshamaseen:
@dcblogdev  fixed.

Two points to take into consideration:
1. I didn't find the lower case that you mentioned here:

> In the module service provider while this does work I think it should be using ucwords its currently:

I've tested all the generates, and they all generate Products/Tags in ucwords.

2. the change in this PR will cause a break for those who are using custom stubs, as some replacements names have been changed to `BASE_*`

mshamaseen:
@dcblogdev I hope this PR doesn't go stale.

h6585:
@dcblogdev I think this feature is nice. However, the language is a confusing factor. I think it should be called as **subfolder** rather than _submodule_

@mshamaseen I am not sure if its a new feature suggested by you? or if its an existing feature or if the PR has already been included. I recently installed this. The version that I have is 8.3 and the ability to make modules located in subfolders is already there. I can make a module inside a folder by using  _php artisan module:make Test/Test_ 

But how do I make controllers, migration, and model files for subfolders and sub-sub folders?

dcblogdev:
thanks this all looks fine now, great work 👏

dcblogdev:
> @dcblogdev I think this feature is nice. However, the language is a confusing factor. I think it should be called as **subfolder** rather than _submodule_
> 
> @mshamaseen I am not sure if its a new feature suggested by you? or if its an existing feature or if the PR has already been included. I recently installed this. The version that I have is 8.3 and the ability to make modules located in subfolders is already there. I can make a module inside a folder by using _php artisan module:make Test/Test_
> 
> But how do I make controllers, migration, and model files for subfolders and sub-sub folders?

it's more than a subfolder as you can place modules into other modules.

I've tried creating modules into subfolders like you described, but it didn't for work me. This PR would make that possible.

When making additional controllers you specify the full module name like 

```
php artisan make:controller PostsController Admin/Posts
```

billypoke:
This feature causes severe degradation in performance in my application after updating from 9.0.1 to 9.0.2. Maybe someone can shed some light. Laravel is the latest version (9.21.6)

```
php -v
PHP 8.1.8 (cli) (built: Jul  8 2022 10:46:35) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.8, Copyright (c) Zend Technologies
    with Xdebug v3.1.4, Copyright (c) 2002-2022, by Derick Rethans
    with Zend OPcache v8.1.8, Copyright (c), by Zend Technologies
```

With `9.0.1`
```
XDEBUG_MODE=off ptest
ParaTest v6.6.1 upon PHPUnit 9.5.21 #StandWithUkraine

...............................................................  63 / 227 ( 27%)
....................................SSSSSSSSSSSS............... 126 / 227 ( 55%)
.....................................................SSSSSSSSSS 189 / 227 ( 83%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS                          227 / 227 (100%)

Time: 00:04.125, Memory: 26.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 227, Assertions: 488, Skipped: 60.
```

With `9.0.2`
```
XDEBUG_MODE=off ptest
ParaTest v6.6.1 upon PHPUnit 9.5.21 #StandWithUkraine

...............................................................  63 / 227 ( 27%)
....................................SSSSSSSSSSSS............... 126 / 227 ( 55%)
.....................................................SSSSSSSSSS 189 / 227 ( 83%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS                          227 / 227 (100%)

Time: 01:08.028, Memory: 26.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 227, Assertions: 488, Skipped: 60.
```
About 15x slowdown 🐌 

Looking at an xdebug profiler snapshot, it looks like a ton of time is being spent scanning files
image

mshamaseen:
@billypoke what is your Structure inside `Modules`? Do you have a lot of folders?

billypoke:
> @billypoke what is your Structure inside `Modules`? Do you have a lot of folders?

277 total directories

find modules/ -type d | wc -l
     277

I have 8 Module directories, each with 5-6 subdirectories, maximum nesting level is 6

mshamaseen:
@billypoke Well, now the package will loop over these directories searching for modules (folders include a direct module.json child), so it is expected to take more time.

Regarding performance, I believe this package is shipped with a caching feature, you can enable caching in the configuration; this should significantly enhance the performance in your case.

billypoke:
@mshamaseen enabling caching does restore the performance. It might be a good idea to change the default behavior to enable caching since this new feature is always enabled and others might run into a similar issue. 

Updating the docs to recommend enabling caching in order to avoid any filesystem performance issues would also be a good idea in my opinion.

rah-emil:
⚠️⚠️⚠️The release broke our app. The whole project did not start! Horror. We stayed on version 9.0.1. Minus 2 days of work)))

mwamodo:
I see the idea of this feature. But it has introduced something that will not be used by a lot of people, and increased complexity overall. 

It broke our app and I suspect its the cause of out app slowing down a whole lot. Maybe it also has something to do with [this boot performance issue](https://github.com/nWidart/laravel-modules/issues/1420)

dcblogdev:
hi @rah-emil broke your app how? do you have any errors in your logs? 

I'm running this on multiple projects the only things I did have to change were on in the scan config `base_path('modulesCms/*/*')` to `base_path('modulesCms') but I think that was specific to my project.

mshamaseen:
@mwamodo 

> But it has introduced something that will not be used by a lot of people

This feature has been requested a LOT of times, but no one introduced a PR for it; see the following:
https://github.com/nWidart/laravel-modules/pull/1186
https://github.com/nWidart/laravel-modules/issues/729
https://github.com/nWidart/laravel-modules/issues/578
https://github.com/nWidart/laravel-modules/issues/201

Now regarding performance, saying "this broke our app" or "it doesn't work" doesn't help us figure out what the issue is; please provide a complete description of what you are facing, including how many modules you have, how many submodules, what exactly happened and what errors did you get, and do you have caching enabled or not.

I'm aware that his feature **WILL** make a performance issue **if** 1. You have a lot of modules, and 2. caching is not enabled. and this is not a bug, you should enable the cache on production.

Sophist-UK:
I am sorry that performance issues meant that this PR was reverted. I think this functionality is useful, but IMO there needs to be an alternative way of finding / registering modules than recursive iteration of a (possibly huge) directory tree. Caching does not stop the tree being searched (which could take literally minutes or even hours on a huge project), it just reduces the frequency.

I appreciate that at present modules are found by a (non-recursive) search of directories, but IMO this needs to be replaced by a means of modules registering themselves in the same way that packages register themselves, rather than LM searching for them.

mshamaseen:
@Sophist-UK I've made a new [PR](https://github.com/nWidart/laravel-modules/pull/1427) to fix the recursive loop issue.

Now even though this new PR stops the use of recursive search, the issue of calling the scan multiple times is still present, and it was present in v9.0 (I am NOT trying to allocate blame, I am just trying to identify if the submodule PR made this issue or not).

To enhance this even better, we need the following:
1. A way to test the performance regardless of the env; I can see from this issue https://github.com/nWidart/laravel-modules/issues/1420 that you have 5 seconds booting the app, whereas I tried it on my device, and I had it booted in 0.7s, maybe it doesn't happen on my side because of a config I changed? Anyway, you know it is tough to fix an issue that doesn't happen in your env.
2. I still can't understand why not use `modules_statuses.json` to get the module paths instead of searching for them every time? is there any reason for that?

Sophist-UK:
1. The issue is happening in your environment, just not enough for you to notice it. But your PRs need to work not just for you but for everyone. Try installing some prerequisites in your modules rather than in your project (i.e. by running `composer require` in the module directory instead.
2. Because `module_statuses.json` is populated from the various `module.json` files found during the recursive search.

Sophist-UK:
I am thinking of starting afresh and switching to `InterNACHI/modular` instead because that supports the normal Laravel package discovery (i.e. the package is registered by composer at install time rather than searched for on each boot of Laravel).

desplega:
Hi!
As of 10th of April 2024, what is the status of this request? Did the submodule functionality finally work?
I can create submodules with "php artisan module:make Test1/Test2", but afterwards the submodule cannot be found. It seems the "slash" issue is not fixed yet, so I guess the feature was finally staled. Am I right?

PR #1418, opened 2022-07-23
Allow symlinks when traversing directories

dcanaud:
Our modules live in a separate git repository and a main laravel project just pulls them in. When in local development, we use symlinks instead of the git repositories to have the changes in the main project immediately instead of doing a git push and a composer update. The recursiveSearch function doesn't consider symlinks when searching for files so the modules are not discovered anymore when we have them symlinked in local. Added the flags to include traversal with symlinks and to ignore folders when traversing.


src/FileRepository.php
```diff
@@ -147,8 +147,20 @@ public function recursiveSearch(string $folder, string $filename): array
             return [];
         }
 
-        $dir      = new \RecursiveDirectoryIterator($folder);
-        $ite      = new \RecursiveIteratorIterator($dir);
+        $ignoreFiles = function ($fileInfo) {
```
dcblogdev, line 150: I think it would be better if these values could come from a config, not everyone will use jetbrains like vscode users.

I don't understand why this is needed, it appears if the files are in the array or not then return true is still triggered?

dcanaud: Yes, we can put it in a config. It is needed because our Modules are a git repository of their own and has a .git folder, and we need to exclude this from the traversal or else there will be a performance degradation since the .git folder contains a lot of files

PR #1419, opened 2022-07-23
Support a hierarchy of config files

Sophist-UK:
1. Config/package.php
2. Modules/module1/Config/package.php
3. vendor/package/config/config.php


src/Laravel/trait moduleConfigMerge.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Extend Laravel's ServiceProvider to handle config file hierarchies  
+ * 
+ * If a module has required packages, you m,ight have a hierarchy of
```
dcblogdev, line 5: there's an extra, in the word might here.

dcblogdev:
Can you explain the purpose of this PR, I've read configs are merged but I don't understand the benefit, can you explain?

Also, can you run the tests and update the snapshots by running:

```
vendor/bin/phpunit -d --update-snapshots
```

You may need to run this twice before the tests work again.

Sophist-UK:
> Can you explain the purpose of this PR, I've read configs are merged but I don't understand the benefit, can you explain?

Modules can have their own dependencies in composer.json (in this case debugbar). And it may be that the module wants to change the dependency defaults included in a package config file, and so the config files for dependencies (`vendor/.../debugbar/config/config.php`) might be published to the module config directory (as `modules/module/config/debugbar.php`). Then the user might wish to override the model config by copying the config file to `/config/debugbar.php`.

It is not going to be clear whether debugbar ServiceProvider is registered/booted before or after the module - the debugbar merges its default config file into the app config file. The module ServiceProvider needs to ensure that the app config overrides the module config overrides the package default config - but it has no idea whether the service provider has already merged the defaults.

The only way to resolve this is to merge the existing settings first with the module config and then with the app config if it exists - and this is what this trait does.

I will also take a look at the tests and the typo this evening.

P.S. I also use Wikimedia's Composer Merge Plugin to make composer include these dependencies for the project as a whole, however this is not a prerequisite for this functionality which is based on app/module/package hierarchy of config application.

Sophist-UK:
> Also, can you run the tests and update the snapshots by running:
> 
> ```
> vendor/bin/phpunit -d --update-snapshots
> ```

Sorry - I tried to run these against my git clone but I have no idea how to make it work. I tried replacing the version in my project dev environment with the clone and run phpunit there, but it didn't run the tests for LM. phpunit doesn't autoload LM from the git clone directory if I try to run it from there.

Sophist-UK:
Ok - I looked in the Github workflows and copied the commands to run unittests. And at least they were running the tests.

But on both master branch and this branch I got a whole bunch of errors entirely unrelated to this PR - I have no idea how to fix these.

Sophist-UK:
@dcblogdev David - Can you explain why this PR was closed unmerged please? If it is because you don't understand the requirement, I would be happy to give further explanation, otherwise if it was a consequence of the recent force push, I am happy to resubmit it.

dcblogdev:
can you re-submit I closed it when reverting

Issue #1420, opened 2022-07-24
[Major] Boot performance


Sophist-UK:
I am just starting a new Laravel project.

I had a false start without using Laravel Modules, and that would give a response in 1s-2s or even faster on my dev environment.

I then decided that I would use Laravel Modules and started with a fresh Laravel environment.

Now, without changing Apache or php, the boot time is c. 5s. (This is true from the command line as well as from a browser.)

Removing Laravel Modules makes it fast, adding it again makes it slow. I have no idea what the cause is yet. I will take a look and see if I can work out what is causing this and how to fix it, but I am not confident I can work this out.

Sophist-UK:
Ok - I have found several potential causes in `/src/FileRepository.php`:

1.  The `scan` function is called several (13) times to find `module.json` when I run `php artisan` with no parameters, and each time it recurses the same directories. Note: This is only recursing my modules directory and not the vendor tree - if I had `scan.enabled` in `config/modules.php` this would be far far worse.

  IMO, the `scan` function should cache the results of the first scan in a class variable and reuse that result.

2. The `scan.exclude` config value that is coded in this class is not documented, but the code suggests that you can only exclude specific filenames and not paths. It would be better IMO for the exclude functionality to use  `glob` style path/name matching patterns for exclusion.

3. Assuming that both `recursiveSearch`  and `exclude` use specific filenames, we do not check once at the beginning of the function if the file we are searching for is in the exclude list and return immediately - instead we recurse all directories and filter out such files which seems to me to be a waste of time. (If we use `glob` style filters, then it would still need to be done in the `RecursiveCallbackFilterIterator`.)

4. The callback gets the `scan.exclude` array on every iteration rather than once when the function is called. 

5. The `recursiveSearch` function is looping through every directory and then every file and getting the filename and comparing it to the wanted filename. It would be much better to iterate through the directories and use an improved `RecursiveCallbackFilterIterator` callback function to check that the filename is correct or indeed only iterate the directories and for each directory check whether the specific file exists.

I will try and see if I can address these issues and create a PR tomorrow as it is time for bed here in PT.

codeanddeploy:
Up for this, I hope it gets fixed.

mshamaseen:
@Sophist-UK There is a caching feature that you can enable in the config file.

I like point 5 of your comment. Can you make a PR for it?

Does enabling the cache in the config file fix it for you? How many folders do you have in your modules folder?

Sophist-UK:
[...]

PR #1421, opened 2022-07-26
Improve FileRepository performance

Sophist-UK:
See #1420 for issue.

This PR addresses points 3., 4. and 5. in the reported issue.

> 3. Assuming that both `recursiveSearch`  and `exclude` use specific filenames, we do not check once at the beginning of the function if the file we are searching for is in the exclude list and return immediately - instead we recurse all directories and filter out such files which seems to me to be a waste of time. (If we use `glob` style filters, then it would still need to be done in the `RecursiveCallbackFilterIterator`.)
>
> 4. The callback gets the `scan.exclude` array on every iteration rather than once when the function is called. 
>
> 5. The `recursiveSearch` function is looping through every directory and then every file and getting the filename and comparing it to the wanted filename. It would be much better to iterate through the directories and use an improved `RecursiveCallbackFilterIterator` callback function to check that the filename is correct or indeed only iterate the directories and for each directory check whether the specific file exists.

The significant performance benefits arise from point 5.

### DIFF
--- a/src/FileRepository.php
+++ b/src/FileRepository.php
@@ -147,23 +147,23 @@ abstract class FileRepository implements RepositoryInterface, Countable
             return [];
         }
 
-        $ignoreFiles = function ($fileInfo) {
-            if (in_array($fileInfo->getFilename(), $this->config('scan.exclude', []))) {
-                return false;
-            }
+        if (in_array($filename, $this->config('scan.exclude', []))) {
+            return [];
+        }
 
-            return true;
-        };
-
-        $dir      = new \RecursiveDirectoryIterator($folder, \FilesystemIterator::KEY_AS_PATHNAME|\FilesystemIterator::CURRENT_AS_FILEINFO|\FilesystemIterator::FOLLOW_SYMLINKS);
-        $filtered = new \RecursiveCallbackFilterIterator($dir, $ignoreFiles);
-        $ite      = new \RecursiveIteratorIterator($filtered);
+        // Iterate directories
+        $dirs = new \FilesystemIterator($folder, 
+            \FilesystemIterator::KEY_AS_PATHNAME
+            | \FilesystemIterator::CURRENT_AS_FILEINFO 
+            | \FilesystemIterator::SKIP_DOTS 
+            | \FilesystemIterator::FOLLOW_SYMLINKS
+        );
 
         $fileList = [];
-        foreach ($ite as $file) {
-            /** @var SplFileInfo $file */
-            if ($file->getFilename() === $filename) {
-                $fileList[] = $file->getPathname();
+        foreach ($dirs as $dir) {
+            $Pathname = $dir->getPathname() . '/' . $filename;
+            if (is_file($Pathname)) {
+                $fileList[] = $Pathname;
             }
         }
 


Sophist-UK:
Annecdotal performance measurements on a (relatively) new Laravel project with two newly created modules as measured by debugbar.

Current Laravel Modules version from packagist - Boot time: 7.52s (yes - over 7 seconds on an i7 with SSD!!!!!)
image

Code in this PR - Boot time: 142ms (yes, really - over 50x speed improvement overall)
image

Literally the only difference is changing the one file.

Sophist-UK:
I will take a look at why the unit tests are failing.

Sophist-UK:
By following the commands in the Github actions, I have worked out how to run unit tests but on master they fail with hundreds of messages like this:
```
1) Nwidart\Modules\Tests\Activators\FileActivatorTest::it_creates_valid_json_file_after_enabling
Error: Class "Modules\WrongModule\Providers\WrongModuleServiceProvider" not found
```

Until I can make phpunit work on my clone, I cannot work out what is breaking in this PR branch.

mshamaseen:
I'm checking the performance on the feature PR commit of [9fe15aecd72d0767ff254a3ec66495be3bdd19c8](https://github.com/nWidart/laravel-modules/commit/9fe15aecd72d0767ff254a3ec66495be3bdd19c8), and it boots with less than a second even with five modules/submodules exist.

7 seconds (as you mentioned) is an insane time, but looks like this happened because of a change after the feature PR.
image

Sophist-UK:
I am not trying to allocate blame - just fix the problem I was experiencing with the 9.0.2 version of the code.

The version of the code in this PR is the fastest possible way of recursing the directories to look for the required file.

 I just need to get the unit tests working (from master on my PC - any help gratefully received) so I can fix the errors reported i github actions and make this ready for merging.

dcblogdev:
if you have access to GitHub codespaces you can open this up in there, I've recorded a quick video to show the process https://www.loom.com/share/5eaeaaffb36b480299b380c23d631669

I appreciate the help you're giving to resolve this issue.

Sophist-UK:
@dcblogdev David

Sorry - I don't have access to code spaces.

I have a local environment, and can run `composer test` or `vendor\phpunit\phpunit\phpunit` but whatever I do I keep getting `Error: Class "Modules\WrongModule\Providers\WrongModuleServiceProvider" not found` for almost every test case.

dcblogdev:
strange I'm not getting that here's what I did https://www.loom.com/share/dfb902359c764462b90b81e3a421b721

Sophist-UK:
@dcblogdev David

Sorry - but that merge was really **NOT** helpful. The code was broken (as it didn't recurse) so now we have a broken laravel-modules. **Can you please revert that merge!!!**

I have made the recursion work using `RecursiveDirectoryIterator` and `ParentIterator` to limit the iterator to just directories and then checking each directory for the specific file we want, but this is not much faster than the problematic version . The reality is that scanning a directory tree is time-consuming, made worse by it being done repeatedly (13x) rather than reusing the results previously found. The existing caching functionality might avoid these 13x and indeed might even avoid the first time for each Laravel boot.

It is also made worse by my using a recursive composer plugin to allow modules to have their own composer.json requirements that get merged together - because this creates large vendor subdirectories elsewhere. (The overhead of composer recursing the directory is a one-off off-line process rather than something that happens on every initialisation of Laravel - and so much less of a performance issue.)

But remember that my project is still a relatively empty structure - but it is intended to be a huge project, so for me I think this could get at least 2 orders of magnitude worse, so any recursive search of directories is going to be a problem in due course.

I have two mitigation strategies in mind:

1. Have a modules config string defining whether sub-modules are allowed inside normal modules, and whether `/vendor` or `/modules/**/vendor` should be searched for other sub-modules

2. Avoid doing this 13x. 

Stepping back a bit, I actually think that searching for the modules is the wrong strategy in any case. Laravel deliberately avoids this for packages by automatically allowing them to register their ServiceProviders in `/config/app.php` when the package is installed - and it seems to me that either:

a. Laravel Modules should adopt the same approach i.e having the modules's service providers added to `config/app.php`; or

b. Instead as a module is autoloaded by PSR4 it registers its existence and then the Laravel Modules boot / register functions would iterate through the list of modules and load each of the Laravel ServiceProviders. However, this feels like it might go against the flow (i.e. by introducing new global variables when the whole direction of PHP is to avoid them).

If we don't already do so, we should also allow deferred bindings for modules like they are allowed for packages.

dcblogdev:
yeah I'm going to roll back before the sub modules was added, that will need looking at but it cannot such a performance problem as you've listed above and through benchmarking.

Unfortunately this does mean the changes over the last few days will be removed.

Sophist-UK:
@dcblogdev David

> strange I'm not getting that here's what I did https://www.loom.com/share/dfb902359c764462b90b81e3a421b721

Thanks for these videos - you are not to know this, but whilst I am new to Composer and Laravel, I have been doing open source contributions via Github and Github Desktop for years. So I have cloned and downloaded Laravel Modules into GH Desktop and run `composer install` but when I run phpunit I get the funny errors as posted above.

Given that I am going to have to rethink the whole thing, I will delete my local clone (and installed pre-requisites) and download and reinstall pre-reqs again.

Sophist-UK:
> yeah I'm going to roll back before the sub modules was added, that will need looking at but it cannot such a performance problem as you've listed above and through benchmarking.
> 
> Unfortunately this does mean the changes over the last few days will be removed.

Given that the sub-modules stuff is relatively new (6 days?), I don't think it will be a major issue to roll it back.

But there is a lesson to be learned here - which is to be sceptical about merging things until you are as certain as possible that there will not be unintended consequences.

dcblogdev:
indeed, needs to do more checking when such large changes are proposed.

Sophist-UK:
@dcblogdev David

I deleted the local clone in GH Desktop, redownloaded it, ran `composer verify`, `composer install` and either `php vendor\phpunit\phpunit\phpunit` or `compose test` and had the same issues. I am stumped.

Could it be a Windows vs Linux issue?

dcblogdev:
I haven't reverted it yet, I'm making sure everything is back to where it was previously I'll then push them up.

Sophist-UK:
You should easily be able to revert commits in reverse sequence in GH Desktop.

dcblogdev:
I tried using GH Desktop it could revert but then couldn't push up as it needed to be forced due to deleting commits.

I end up doing it manually. easier than trying to fight with GUI tools.

PR #1428, opened 2022-07-28
Feature/allow submodule

mshamaseen:
This is a fix for the performance issue of https://github.com/nWidart/laravel-modules/pull/1409

This allows the developer to customize how many sub-directory to search for module.json by defining that explicitly on the config.scan; by default, it just allows on a directory under Modules (so no sub-modules by default) to allow for more search levels you should enable the scan in the config and explicitly add to the levels like this:

```
'scan' => [
        'enabled' => true,
        'paths' => [
            base_path('Modules/*'),
            base_path('Modules/*/*'),
            base_path('Modules/*/*/*'),
        ],
    ],
```
In this fix, I used glob instead of recursive search, the same as the original, with all the other functionality of creating a submodule being the same.

I also added [GLOB_NOSORT](https://www.phparch.com/2010/04/putting-glob-to-the-test/) as it supposes to enhance the performance.

Please note that the scan function is still called many times; this issue exists even on v9.0, and I haven't fixed it on this PR.


mshamaseen:
It makes me sad that this effort went stale ..

PR #1438, opened 2022-08-13
Update DisableCommand.php

ajayfroiden:
Fix the below issue on run 
`php artisan module:disable`

Nwidart\Modules\FileRepository::findOrFail(): Argument #1 ($name) must be of type string, null given, called in 
/vendor/nwidart/laravel-modules/src/Commands/DisableCommand.php on line 38


dcblogdev:
thanks!

PR #1591, opened 2023-07-12
[feat] update DisableCommand  to get list of modules

aryala7:
this PR adds feature to accept array of modules instead of single module to disable

### DIFF
--- a/src/Commands/DisableCommand.php
+++ b/src/Commands/DisableCommand.php
@@ -16,11 +16,18 @@ class DisableCommand extends Command
     protected $name = 'module:disable';
 
     /**
+     * The console command signature.
+     *
+     * @var string
+     */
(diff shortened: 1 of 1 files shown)


src/Commands/DisableCommand.php
```diff
@@ -29,11 +36,10 @@ public function handle(): int
     {
         $this->components->info('Disabling module ...');
 
-        if ($name = $this->argument('module') ) {
+        foreach($this->argument('module') as $name) {
             $this->disable($name);
-
-            return 0;
         }
+        return 0;
```
CrazyTapok-bit, line 42: This is incorrect, because the code below will never be executed

aryala7:
@CrazyTapok-bit  i have fixed it.

PR #1601, opened 2023-07-31
[feat] update enable command to also accept array of module names

aryala7:


dcblogdev:
Looks good, but can you ensure you can pass a single module name ie: `php artisan module:enable Blog` in addition to passing an array.

PR #1623, opened 2023-10-03
Add Make Observer Command

hanieas:
After merging this PR we can make a new observer based on the selected model, The related command to be run is:

```
php artisan module:make-observer Post Blog
```

This command will create `PostObserver` inside the `Moudles\Blog\Observers` directory and contain all necessary functions to be observed.


dcblogdev:
thanks!!

PR #1625, opened 2023-10-14
Command $name property works but it doesn't support options…

Rattone:
…and maybe even arguments

I lost more than an hour to realize that the signature was actually the name, and for this reason I couldn't pass options to the command. This was the only command I had generated from this stub and I was punished :-P

### DIFF
--- a/src/Commands/stubs/command.stub
+++ b/src/Commands/stubs/command.stub
@@ -13,7 +13,7 @@ class $CLASS$ extends Command
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


Rattone:
Sorry for not updating the tests. I tried to understand something but I have doubts with snapshots, should I update them or are they generated somehow?

dcblogdev:
for snapshots, you can run tests with `-d --update-snapshots` which then update the snapshots.

I can do this for you.

PR #1628, opened 2023-10-20
Update config.php

solomon-ochepa:
Format the file structure to the current Laravel standard.

These changes will only apply to newly created modules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All Notable changes to `laravel-modules` will be documented in this file.
 
 ## Next
 
+## 10.0.2 - 2023-09-18
+
+## Changed
+
+- reordered config commands and added missing observer command
+
 ## 10.0.1 - 2023-09-18
 
 ## Added
--- a/config/config.php
+++ b/config/config.php
@@ -29,14 +29,14 @@ return [
         'enabled' => false,
         'path' => base_path('vendor/nwidart/laravel-modules/src/Commands/stubs'),
         'files' => [
-            'routes/web' => 'Routes/web.php',
-            'routes/api' => 'Routes/api.php',
-            'views/index' => 'Resources/views/index.blade.php',
-            'views/master' => 'Resources/views/layouts/master.blade.php',
-            'scaffold/config' => 'Config/config.php',
+            'routes/web' => 'routes/web.php',
+            'routes/api' => 'routes/api.php',
+            'views/index' => 'resources/views/index.blade.php',
+            'views/master' => 'resources/views/layouts/master.blade.php',
+            'scaffold/config' => 'config/config.php',
             'composer' => 'composer.json',
-            'assets/js/app' => 'Resources/assets/js/app.js',
-            'assets/sass/app' => 'Resources/assets/sass/app.scss',
+            'assets/js/app' => 'resources/assets/js/app.js',
+            'assets/sass/app' => 'resources/assets/sass/app.scss',
             'vite' => 'vite.config.js',
             'package' => 'package.json',
         ],
@@ -66,8 +66,8 @@ return [
         | Modules path
         |--------------------------------------------------------------------------
         |
-        | This path used for save the generated module. This path also will be added
-        | automatically to list of scanned folders.
+        | This path is used to save the generated module.
+        | This path will also be added automatically to the list of scanned folders.
         |
         */
 
@@ -77,17 +77,17 @@ return [
         | Modules assets path
         |--------------------------------------------------------------------------
         |
-        | Here you may update the modules assets path.
+        | Here you may update the modules' assets path.
         |
         */
 
         'assets' => public_path('modules'),
         /*
         |--------------------------------------------------------------------------
-        | The migrations path
+        | The migrations' path
         |--------------------------------------------------------------------------
         |
-        | Where you run 'module:publish-migration' command, where do you publish the
+        | Where you run the 'module:publish-migration' command, where do you publish the
         | the migration files?
         |
         */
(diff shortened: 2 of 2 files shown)


solomon-ochepa:
Please, I need an explanation as to why this PR was closed without being merged.

This particular PR is crucial, and I expected it to be treated before any other. Please quote me if I'm wrong!

dcblogdev:
this is making sweeping changes that I don't want. This would require a version dump due to breaking changes.

In future, before making fundamental changes I suggest asking to see if it would be well received to save your time.

You're free to change the config to your liking in your application. I try to keep changes in the config to essential changes only to avoid breaking existing applications that may not have published the config file.

solomon-ochepa:
> this is making sweeping changes that I don't want. This would require a version dump due to breaking changes.
> 
> In future, before making fundamental changes I suggest asking to see if it would be well received to save your time.
> 
> You're free to change the config to your liking in your application. I try to keep changes in the config to essential changes only to avoid breaking existing applications that may not have published the config file.

You are right, but not totally right.

```json
'statuses-file' => base_path('modules.json'),
```
There is a breaking change at `line: 274`, which is true (sorry, my mistake).

However, the rest of the code is as clean as possible, and wouldn't cause any form of breaking changes from Laravel 5 up to date.

Please, do point out any of other observations.

Thanks.

solomon-ochepa:
Please, can you reopen the PR so I can make an edit so we can continue the review?

dcblogdev:
You're right the generator changes don't break existing modules. I don't think its a good idea to be a mix of namespaces names ie:

a controller would have lower case app

```
namespace Modules\Contacts\app\Http\Controllers;
```

or a seeder

```
namespace Modules\Contacts\database\seeders;
```

They would be better to match the style ie

```
namespace Modules\Contacts\App\Http\Controllers;
namespace Modules\Contacts\Database\Seeders;
```

solomon-ochepa:
> 
> They would be better to match the style ie
> 
> ```
> namespace Modules\Contacts\App\Http\Controllers;
> namespace Modules\Contacts\Database\Seeders;
> ```
> 
> 
> 

That's completely wrong, following the Laravel code standard.

Note that the namespace should be assigned in accordance with the path! Note that the app directory is in lower case.

dcblogdev:
I disagree, for instance, looking at seeders from Laravel their namespace is `Database\Seeders` not `database\seeders`

solomon-ochepa:
> I disagree, for instance, looking at seeders from Laravel their namespace is `Database\Seeders` not `database\seeders`

That's true, that's manipulated via the composer autoloader. Laravel modules on the other hand only loaded the `Modules/`

Notwithstanding, I'd like you to take a look at your newly created Laravel app. Observe the `/database/migrations`, `/database/seeders` paths. Following the naming convention of the paths, the namespace should match the case else it will throw error.

You may try the both cases in your local host.

Thanks.

solomon-ochepa:
Hi, [dcblogdev](https://github.com/dcblogdev).

I'm still waiting for you here, please.

dcblogdev:
if you can get this to work with the namespaces `Modules\Contacts\Http\Controllers` etc we can move forward.

I've tried the config and adding 

```
"psr-4": {
    "Modules\\Books\\": "app/",
    "Modules\\Books\\Tests\\": "tests",
    "Modules\\Books\\Database\\Factories\\": "database/factories",
    "Modules\\Books\\Database\\Seeders\\": "database/seeders"
}
```

but having problems getting this to work.

solomon-ochepa:
> If you can get this to work with the namespaces `Modules\Books\Http\Controllers` etc we can move forward.
> 
> I've tried the config and adding
> 
> ```
> "psr-4": {
>     "Modules\\Books\\": "app/",
>     "Modules\\Books\\Tests\\": "tests",
>     "Modules\\Books\\Database\\Factories\\": "database/factories",
>     "Modules\\Books\\Database\\Seeders\\": "database/seeders"
> }
> ```
> 
> but having problems getting this to work.

You are complicating the whole process! What I worked on is not the composer.json file within each module as in your example above (maybe someone can come up with that solution in the future).

This PR aims to give each new module the same file structure as the latest Laravel version.

From Laravel v5.* to the current v10.*, the Controller namespace, for instance, has been `app/Http/Controllers`.

So, if a new module `Book` is to be created using the latest Laravel version, it would be Modules\Books\ `app\Http\Controllers, instead of Modules\Books\ `Http\Controllers` as given in your example above.

Making the `app` in the namespace appear as `App` is not the aim of this PR in the first place, so where is your argument coming from?

We can work on making the `app` appear as `App` in subsequent PR, but not within the scope of this particular PR, in the sense that the `/composer.json` responsible for this PR effect is entirely different from the `/Modules/Books/composer.json,` responsible for the result you are seeking. Do you get my point now?

solomon-ochepa:
> ...
> I've tried the config and adding
> 
> ```
> "psr-4": {
>     "Modules\\Books\\": "app/",
>     "Modules\\Books\\Tests\\": "tests",
>     "Modules\\Books\\Database\\Factories\\": "database/factories",
>     "Modules\\Books\\Database\\Seeders\\": "database/seeders"
> }
> ```
> 
> but having problems getting this to work.

Please take note of the `"Modules\\Books\\": "app/"` it should be `"Modules\\Books\\App": "app/"` instead.

Try this out instead:
```
"psr-4": {
    "Modules\\Books\\App\\": "app/",
    "Modules\\Books\\Tests\\": "tests",
    "Modules\\Books\\Database\\Factories\\": "database/factories",
    "Modules\\Books\\Database\\Seeders\\": "database/seeders"
}
```

solomon-ochepa (commented):
The namespace can be changed from the config file.

Notwithstanding, I'm working on how to get the value directly from the module's `composer.json` file.

mstojanovv:
This change has resulted in a less appealing namespace structure, in fact, not really following Laravel modern architecture or standards.

The usage of lowercase directory namespace like `Modules\ModuleName\app\Providers`, `Modules\ModuleName\database\seeders`, affects the overall code.

I believe that @dcblogdev or @solomon-ochepa  should have addressed the issue of these lowercase directory related namespace, including **app**, **database**, and any others located at the root of the module, before merging the change into the package. 

The PR is great, but using a lowercase namespace folder is not that much.

Despite my attempts to find various solutions, I haven't found any solution to make it working.

solomon-ochepa:
> This change has resulted in a less appealing namespace structure, in fact, not really following Laravel modern architecture or standards.
> 
> The usage of lowercase directory namespace like `Modules\ModuleName\app\Providers`, `Modules\ModuleName\database\seeders`, affects the overall code.
> 
> I believe that @dcblogdev or @solomon-ochepa  should have addressed the issue of these lowercase directory related namespace, including **app**, **database**, and any others located at the root of the module, before merging the change into the package. 
> 
> The PR is great, but using a lowercase namespace folder is not that much.
> 
> Despite my attempts to find various solutions, I haven't found any solution to make it working.

Point of correction, the structures are accurate and follows Laravel current standard.

The only point we are currently missing and seeking solution to is how to map the namespaces.

Like in Laravel, the `app/`, `database/migrations/` & `database/seeders` namespaces are not based on the paths lowercase names, but rather were config via the `composer.json` file.

We are currently working towards that, in short we have added the necessary properties to the Module's `composer.json` already.

So, please, this issue has nothing to do with the lowercase directories.

PR #1650, opened 2023-10-21
Update controller-api.stub

solomon-ochepa:
Formatted.


solomon-ochepa:
If I may ask, why was this PR closed without being merged?

We need an explanation, please.

PR #1655, opened 2023-10-21
Update api.stub

solomon-ochepa:
Formatted.

### DIFF
--- a/src/Commands/stubs/routes/api.stub
+++ b/src/Commands/stubs/routes/api.stub
@@ -1,18 +1,19 @@
 <?php
 
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Route;
 
 /*
-|--------------------------------------------------------------------------
-| API Routes
-|--------------------------------------------------------------------------
-|
-| Here is where you can register API routes for your application. These
-| routes are loaded by the RouteServiceProvider within a group which
-| is assigned the "api" middleware group. Enjoy building your API!
-|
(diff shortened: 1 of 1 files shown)


solomon-ochepa:
If I may ask, why was this PR closed without being merged?

dcblogdev:
there is not a v1 group out the box, it would be better to the namespace but omit the prefix

solomon-ochepa:
> there is not a v1 group out the box, it would be better to the namespace but omit the prefix

The `v1` is a versioning system, which is recommended and a standard for all API resources.

solomon-ochepa:
> > there is not a v1 group out the box, it would be better to the namespace but omit the prefix
> 
> The `v1` is a versioning system, which is recommended and a standard for all API resources.

API responses may change from time to time, and the best way to keep track is with the versioning system; `v1`, `v2`, `v3`, etc.

This feature might not be built into this package from the beginning, but that doesn't mean standard can't be recommended.

Thanks.

dcblogdev:
I agree, I was thinking that the versions are coded into the route provider, which they're not I forgot as that's how I normally structure my API's.

solomon-ochepa:
> there is not a v1 group out the box, it would be better to the namespace but omit the prefix

Please, before you close any PR next time, always think about the needs of others.

The fact that a PR doesn't apply to your specific needs or expectations doesn't mean the community doesn't need it. Always think, putting the community first.

PR #1656, opened 2023-10-21
Update web.stub

solomon-ochepa:
Apply route resources and support for soft-delete


dcblogdev:
modules out the box does not have an auth system.

solomon-ochepa:
> modules out the box does not have an auth system.

You are right, sorry I'll make an amend and resubmit.

solomon-ochepa:
Can you please reopen this PR so I can make the necessary amendments and we can continue the review?

Thanks.

PR #1669, opened 2023-10-30
Update master.stub

solomon-ochepa:
Update the Layout using the latest Laravel Layout (`resources/views/layouts/guest.blade.php`) format and codebase.

A few other extra meta tags have been added for SEO purposes.

### DIFF
--- a/src/Commands/stubs/views/master.stub
+++ b/src/Commands/stubs/views/master.stub
@@ -1,19 +1,29 @@
 <!DOCTYPE html>
-<html lang="en">
-    <head>
-        <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


dcblogdev:
you can remove the keywords search engines don't use them.

regarding the use of slots either add in a yield as well so components or extends can be used or the view index stub will need to be updated to use the layout as a component

solomon-ochepa:
> or the view index stub will need to be updated to use the layout as a component

That is very correct. I'm updating it right away.

Concerning the `keywords` meta, we all know that SEO doesn't care about it for the moment, but it's recommended, not just for SEO only.

solomon-ochepa:
> regarding the use of slots either add in a yield as well so components or extends can be used or the view index stub will need to be updated to use the layout as a component

https://github.com/nWidart/laravel-modules/pull/1676

PR #1696, opened 2023-12-03
Module EventServiceProvider

solomon-ochepa:
Create an `EventServiceProvider` class on module creation and an artisan command `php artisan module:EventServiceProvider Blog` to create a new `EventServiceProvider` for an existing module.

* Auto registered.
* Auto-discovery enabled by default.
* Enables module's Observers and Events to be registered locally instead of using the `App\Providers\EventServiceProvider`


dcblogdev:
can you fix the failing tests. 

Also, EventServiceProvider should not be created by default instead use the new command to generate them as needed. No benefit in creating a provider when in most cases it won't be used.

Better to generate the provider when its needed.

dcblogdev:
> No events or observers will work without this file.

yes so generate an events module when you need to use events or observers otherwise its simply not needed.

solomon-ochepa:
I need to create an Event or Observer within any Module and checked if it will work.

This package is a solution, and of what essence is it if it requires a complete work around to make included features (Events, Observers, etc.) function.

Please, note that this also require modifying the package core code if a user are to fix it themselves. Is it recommended to edit the package core code to enable a feature?

dcblogdev:
The reason you've created the module:EventServiceProvider command is to generate the provider, surely you can use that when needed. 

Also the command would be better to be renamed to 'module:event-provider` to match the `module:route-provider` style.

Every module does not need an events service provider. I've worked on systems with around 20 modules only a couple had event service providers since they used observers the rest did not and still work absolutely fine.

dcblogdev:
okay a misunderstanding, lets move on.

My understanding is that the PR is a command to generate module EventServiceProviders. 

You can use that command to generate EventServiceProviders for modules that need it. Help me to understand why you cannot use that?

dcblogdev:
> What about we add an option to the config to allow dev choose when to enable or disable the auto creation feature? I believe that will solve the problem, right?

yes, that would be ideal.

PR #1730, opened 2024-02-01
Fix for Symfony/Console

jyxjjj:
vendor/symfony/console/Input/InputOption.php:85


jyxjjj:
Original can run with Symfony/Console=5.4.23 but cannot run with 5.4.35.
I have not tested for between versions.

This commit should be named as nWidart/laravel-modules=8.3.1 and newer.

PR #1736, opened 2024-02-06
delete command `module:migrate-fresh`

alissn:
Hi,

The `migrate:fresh` command in Laravel fully supports this functionality. This command executes `migrate:fresh`, followed by `module:migrate` (not need).


dcblogdev:
sorry, I'm not sure I understand?

alissn:
@dcblogdev 
The command `module:migrate-fresh` first executes `migrate:fresh`, which entails dropping all tables and then migrating all migrations from the root project and all module migrations.

Following that, it runs `module:migrate` with the argument `module`. The result is:

```sh
   INFO  Nothing to migrate.
```

On the other hand, all arguments that exist in `migrate:fresh` are also present in `module:migrate`.

Module Migrate Fresh Screenshot

dcblogdev:
okay yes, that makes sense now. I'll merge it as part of the v11 release cycle.

src/Commands/MigrateFreshCommand.php
```diff
@@ -1,91 +0,0 @@
...
-    {
-        $module = $this->argument('module');
-
-        if ($module && !$this->getModuleName()) {
-            $this->error("Module [$module] does not exists.");
-
-            return E_ERROR;
-        }
-
-        $this->call('migrate:fresh');
```
solomon-ochepa, line 41: This command should be enhanced to affect only the specified module. Instead of dropping the feature entirely!

For instance, if a developer executes the `php artisan module:migrate-fresh User`, it should refresh migrations from the User module exclusively. If no module is provided, the command should execute for all modules.

src/Commands/MigrateFreshCommand.php
```diff
@@ -1,91 +0,0 @@
...
-
-        $this->call('module:migrate', [
-            'module' => $this->getModuleName(),
-            '--database' => $this->option('database'),
-            '--force' => $this->option('force'),
-            '--seed' => $this->option('seed'),
-        ]);
```
solomon-ochepa, line 48: Certainly, this section should be removed.

After running the command (`module:migrate-fresh`), the dev can choose to run this second command (`module:migrate`) separately.

solomon-ochepa (changes requested):
This feature is necessary, except if it can't be achieved within the scope of the module.

I recommend that we retain the command and focus on improving its functionality instead.

alissn:
@solomon-ochepa

Hi! 
The command `module:migrate-refresh` does exactly what you want. 

First, it runs `module:migrate-reset`, then it calls `module:migrate`, and finally, it executes `module:seed` for specific modules.

solomon-ochepa:
> @solomon-ochepa
> 
> Hi! 
> The command `module:migrate-refresh` does exactly what you want. 
> 
> First, it runs `module:migrate-reset`, then it calls `module:migrate`, and finally, it executes `module:seed` for specific modules.
> 

That's 3 commands actually. This command should focus on refreshing the migrations alone.

I understand it's a nice idea combining the three commands together, but every commands should be isolated.

solomon-ochepa:
module:migrate-refresh
module:migrate
module:seed

The three commands should not be combined by default.

If it's important that you call the three together, the options/flags can be added such as --seed --migrate

alissn:
@dcblogdev 
Did you take a look at this file? This command exists in the package.

You can find it here: https://github.com/nWidart/laravel-modules/blob/master/src/Commands/MigrateRefreshCommand.php

solomon-ochepa:
> @dcblogdev 
> Did you take a look at this file? This command exists in the package.
> 
> You can find it here: https://github.com/nWidart/laravel-modules/blob/master/src/Commands/MigrateRefreshCommand.php
> 

The flows in this file are more accurate and understanding.

Seem like the refresh will not work after all. So, it's first reset, then migrated which is equivalent to refresh. Then seeded if the --seed flag is provided.

This is all we wanted!

solomon-ochepa:
> @dcblogdev 
> Did you take a look at this file? This command exists in the package.
> 
> You can find it here: https://github.com/nWidart/laravel-modules/blob/master/src/Commands/MigrateRefreshCommand.php
> 

That is the same file you were asking to be removed.

I believe we can close this PR now and maintain the updated file.

alissn:
> That is the same file you were asking to be removed.

this PR delete `src/Commands/MigrateFreshCommand.php` file.
it's different with `src/Commands/MigrateRefreshCommand.php`

solomon-ochepa:
> > That is the same file you were asking to be removed.
> 
> this PR delete `src/Commands/MigrateFreshCommand.php` file. it's different with `src/Commands/MigrateRefreshCommand.php`

Both commands should be maintained. There is a reason why even the Laravel framework maintains both commands.

Nevertheless, the `module:migrate-fresh` command should run within the scope of the targeted module only; otherwise it should be updated. And if there is no means, we can proceed with deleting it.

The project aims to replicate the core Laravel structure, commands and workflow in a modular architecture.

dcblogdev:
I think for the same of compatability, we could update `MigrateFreshCommand` to call refresh ie

```
$this->call('module:migrate-refresh', [
    'module' => $this->getModuleName(),
    '--database' => $this->option('database'),
    '--force' => $this->option('force'),
    '--seed' => $this->option('seed'),
]);
```

dcblogdev:
Closed, going to restrict module:migrate-fresh to reset only selected modules this way its only in the scope of the module and not the whole application.

PR #1737, opened 2024-02-07
fix typo

slps970093:
Hello
in the config/config.php file
'resource' => ['path' => 'App/resources', 'generate' => false],
should be
'resource' => ['path' => 'App/Resources', 'generate' => false],


dcblogdev:
this is not a typo, the config option can be changed as desired.

PR #1765, opened 2024-03-17
Solomon ochepa update blade view layout

solomon-ochepa:
Update the view blade to use the modern slot method.

### DIFF
--- a/src/Commands/stubs/views/index.stub
+++ b/src/Commands/stubs/views/index.stub
@@ -1,7 +1,5 @@
-@extends('$LOWER_NAME$::layouts.master')
-
-@section('content')
+<x-app-layout :data="$head ?? []">
     <h1>Hello World</h1>
 
     <p>Module: {!! config('$LOWER_NAME$.name') !!}</p>
-@endsection
+</x-app-layout>
--- a/src/Commands/stubs/views/index.stub
+++ b/src/Commands/stubs/views/index.stub
@@ -1,4 +1,4 @@
-<x-app-layout :data="$head ?? []">
+<x-app-layout>
     <h1>Hello World</h1>
 
     <p>Module: {!! config('$LOWER_NAME$.name') !!}</p>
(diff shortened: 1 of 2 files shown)


dcblogdev:
instead of loading a global layout component a module one should be loaded ie in the module have a `App\View\Components` folder with `AppLayout.php` then the `master.blade.php` should be renamed to `app.layout.php`

The `index.blade.php can then use the module layout:

would need to be stupified of course.

```php
<x-demo::app-layout>

    <h1>Hello World</h1>

    <p>Module: {!! config('demo.name') !!}</p>
    
</x-demo::app-layout>
```

solomon-ochepa:
> instead of loading a global layout component a module one should be loaded ie in the module have a `App\View\Components` folder with `AppLayout.php` then the `master.blade.php` should be renamed to `app.layout.php`
> 
> The `index.blade.php can then use the module layout:
> 
> would need to be stupified of course.
> 
> ```php
> <x-demo::app-layout>
> 
>     <h1>Hello World</h1>
> 
>     <p>Module: {!! config('demo.name') !!}</p>
>     
> </x-demo::app-layout>
> ```

That makes sense!

Nevertheless, it's very rare to have a module with a dedicated layout. The app layout is the base layout for major Laravel projects.

For convenience, I'll update the PR to include its dedicated layout.

Thank you.

dcblogdev:
I use a central layout for my modules but to keep it in line with the default now it makes sense to have it module-specific. Also I you don't get a layout component out of the box with Laravel, you do with starter kits like Breeze.

> For convenience, I'll update the PR to include its dedicated layout.

thank you.

solomon-ochepa:
> I use a central layout for my modules but to keep it in line with the default now it makes sense to have it module-specific. Also I you don't get a layout component out of the box with Laravel, you do with starter kits like Breeze.
> 
> > For convenience, I'll update the PR to include its dedicated layout.
> 
> Thank you.

That is correct 👍

dcblogdev:
closing this, this is a starting point, views / layouts will always need changing based on the needs of the project being used.

PR #1771, opened 2024-03-19
Update api.stub

solomon-ochepa:
Bind the module's resource Controller.


dcblogdev:
bear in mind, this requires replacements adding into the replaces section of the config for API, I'll merge and update the config.

solomon-ochepa:
> bear in mind, this requires replacements adding into the replaces section of the config for API, I'll merge and update the config.

That is very correct. Thanks a lot for the fix.

PR #1786, opened 2024-03-24
Passing the values of vendor, author's name, and author's email in a line during module creation.

enterprime:
Example of usage:
--author-name=value
--author-email=value
--author-vendor=value

All values are optional.

`php artisan module:make --author-name=Alex`

It is useful for modular development to avoid being tied to the system configuration.


dcblogdev:
thanks!

PR #1794, opened 2024-03-27
Update config.php

solomon-ochepa:
Update the following config keys to a more generic phrase

`MODULES_VENDOR_NAME` instead of `MODULES_NAME`,
`MODULES_VENDOR_EMAIL` instead of `MODULES_EMAIL`.


nWidart:
Hi,

Module name and module vendor name are not the same thing in my opinion.
Module name could be "invoice" but the developer is called "john doe".

solomon-ochepa:
> Hi,
> 
> Module name and module vendor name are not the same thing in my opinion.
> Module name could be "invoice" but the developer is called "john doe".

You are correct, and that's the exact reason why this PR was submitted.

Take a look at the code changes again.

The key `MODULE_NAME` is wrong, it should be `MODULE_VENDOR_NAME` instead!

Thanks for noticing the correction.

dcblogdev:
I think `MODULES_AUTHOR_NAME` and `MODULES_AUTHOR_EMAIL` would make more sense, to make it clear its for the author of the module.

dcblogdev:
Also change `MODULES_VENDOR` to `MODULE_VENDOR` so they all match the intention to add details to a module.

Issue #1800, opened 2024-03-31
Need to add two command "make:enum" and "make:class" for the modules


Siam401:
Laravel 11 added enum and class making command "class"  command is useful when creating custom service class for the controller

solomon-ochepa:
Yeah, I read the release documents and certainly, these two pieces are great. We'll incorporate them as soon as possible.

dcblogdev:
I've added a whole bunch of missing commands, make-class is being worked on.

https://github.com/nWidart/laravel-modules/releases/tag/v11.0.8

PR #1802, opened 2024-04-01
Create module:make-class command

solomon-ochepa:
Issue: https://github.com/nWidart/laravel-modules/issues/1800 - Need to add two command "make:enum" and "make:class" for the modules

```
Description:
  Create a new class

Usage:
  module:make:class [options] [--] <name> <module>

Arguments:
  name                  The name of the class
  module                The targeted module

Options:
  -t, --type[=TYPE]     The type of class, e.g. class, service, repository, contract, etc. [default: "service"]
  -p, --plain           Create the class without suffix (type)
  -i, --invokable       Generate a single method, invokable class
  -f, --force           Create the class even if the class already exists
  -h, --help            Display help for the given command. When no command is given display help for the list command
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi|--no-ansi  Force (or disable --no-ansi) ANSI output
  -n, --no-interaction  Do not ask any interactive question
      --env[=ENV]       The environment the command should run under
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```

### Create a service class:
```
php artisan module:make:class Example Example
```

### Create a standard class
```
php artisan module:make:class Example Example -t class
```

### Create a repository class
```
php artisan module:make:class Example Example -t repository
```

### Create a repository class without a suffix
```
php artisan module:make:class Example Example -p -t repository
```


dcblogdev:
This looks good the signature should be `module:make-class` to match the other commands.

This needs to be compatible with the already released config file please remove config changes that are dependant on your other PR.

Also, tests are required for the new command.

Siam401:
yes this command should look like  module:make-class and module:make-enum

dcblogdev:
I'm not sure about having a classes command, I think it is better to have specific make commands ie:

php artisan module:make-enum
php artisan module:make-service 
php artisan module:make-repository

Thoughts?

dcblogdev:
I'm closing this as I don't want a generic classes folder.  Having a folder named classes isn't particularly common or standardized. Laravel follows a convention-over-configuration approach, meaning it encourages developers to adhere to certain conventions for better organization and readability. 

Developers often create additional directories within app to organize their code further. These directories are often named based on the domain of the application or the functionality they contain. For instance, you might have app/Http, app/Services, app/Helpers, app/Repositories, etc.

solomon-ochepa:
> I'm not sure about having a classes command, I think it is better to have specific make commands ie:
> 
> php artisan module:make-enum php artisan module:make-service php artisan module:make-repository
> 
> Thoughts?

Yes, you are correct.

This PR will enable the generation of various types of classes, including enums, services, repositories, and many other custom types.

I intend to establish this class as a foundation for easily generating dedicated class commands such as `module:make-enum`, `module:make-service`, and `module:make-repository` while also allowing for the creation of custom classes.

After this PR is merged, I'll proceed with the remaining commands immediately.

src/Commands/Make/ClassCommand.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    protected $description = 'Create a new class';
+
+    protected $argumentName = "name";
+
+    public function getTemplateContents()
+    {
+        $module = $this->laravel['modules']->findOrFail($this->getModuleName());
+
+        return (new Stub('/app/Classes/Class.stub', [
```
dcblogdev, line 36: the Class.stub should be in the root, as the base path could be src

src/Commands/Make/ClassCommand.php
```diff
@@ -0,0 +1,108 @@
...
+    public function getClass(): string
+    {
+        return $this->getFileName();
+    }
+
+    public function getDefaultNamespace(): string
+    {
+        $type = $this->option('type');
+
+        return config("modules.paths.generator.{$type}.namespace", $this->path_namespace(config("modules.paths.generator.{$type}.path", $this->type_path('app/Classes'))));
```
dcblogdev, line 87: better to no hardcode app/ the folder should be dynamic.

src/Commands/Make/ClassCommand.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    use ModuleCommandTrait;
+
+    /**
+     * The name and signature of the console command.
+     */
+    protected $signature = 'module:make-class
+        {--t|type=class : The type of class, e.g. class, service, repository, contract, etc.}
+        {--p|plain : Create the class without suffix (type)}
+        {--i|invokable : Generate a single method, invokable class}
```
dcblogdev, line 20: Invokable option not yet implemented.

dcblogdev (changes requested):

src/Commands/Make/ClassCommand.php
solomon-ochepa: The location of the stub file doesn't affect where the class will be placed. The final class file will be created in the path specified by the `generator.paths.app`.
I organize the stub file based on the default Laravel structure to help me and others get to the file easily later.

src/Commands/Make/ClassCommand.php
solomon-ochepa: The path/folder is not hand-coded. The part that seems like hand-code is a fallback value, in case every other config fails, which is necessary.

One thing I can do is to fall back to the `generator.paths.app` before gracefully falling back to the Laravel default where every other option has failed.

src/Commands/Make/ClassCommand.php
solomon-ochepa: I agree with you. I had planned to implement it later on.

solomon-ochepa:
> I think it may be better to generate classes into the folder root or the `'app_folder'` folder that way you can set the path as wanted for example:
> 
> ```shell
> php module:make-class Concerns/HasDomain
> php module:make-class Services/UserService
> php module:make-class Actions/UserStoreAction
> php module:make-class Helpers/DateFormat
> ```
> 
> These would be stored in app/src depending on the value of `app_folder`

This is already the default behaviour. Regarding the `paths.app`, it will be solved with the https://github.com/nWidart/laravel-modules/pull/1802#discussion_r1583166968 review.

> Also the make-class command needs tests to be accepted.

I can't run the test with my current Windows system. I'll appreciate it if you or anyone else can create the test.

dcblogdev:
To handle cases there the config doesn't have the class path you can modify your code from:

```php
$config = GenerateConfigReader::read('class');
$path .= $this->type_path($config->getPath()) . '/' . $this->getFileName() . '.php';
```

To 

```php
$filePath = GenerateConfigReader::read('class')->getPath() ?? config('modules.paths.app_folder') . 'Classes';
$path .= $this->type_path($filePath) . '/' . $this->getFileName() . '.php';
```

Can you change your method names to be camelCased rather then snake_case, also there's a few methods that don't have a return type.

Let me know when you feel this is ready and I can look at added the tests

solomon-ochepa:
> To handle cases where the config doesn't have the class path you can modify your code from:
> 
> ```
> $config = GenerateConfigReader::read('class');
> $path .= $this->type_path($config->getPath()) . '/' . $this->getFileName() . '.php';
> ```
> 
> To
> 
> ```
> $filePath = GenerateConfigReader::read('class')->getPath() ?? config('modules.paths.app_folder') . 'Classes';
> $path .= $this->type_path($filePath) . '/' . $this->getFileName() . '.php';
> ```

I've fixed this already.
```php
    public function getDestinationFilePath(): string
    {
        $app_path = GenerateConfigReader::read('class')->getPath() ?? $this->app_path('Classes');

        return $this->path($this->type_path($app_path) . '/' . $this->getFileName() . '.php');
    }
```

I've added a method to get the `paths.app_folder` by default or fallback to `app/` if empty. To use the module base path, we should consider passing '/' as the value of the `paths.app_folder`
```php

    /**
     * Get the app path basename.
     */
    public function app_path(?string $path = null): string
    {
        $config_path = config('modules.paths.app_folder');
        $app_path = strlen($config_path) ? trim($config_path, '/') : 'app';
        $app_path .= strlen($path) ? '/' . $path : '';

        return $this->clean_path($app_path);
    }
```

> Can you change your method names to camelCased rather than snake_case, ...

The snake_case is easier for me to identify and read. I didn't change any existing identifiers to my preference case, as every developer has their preferences. However, the PHP convention supports the snake_case - https://www.php.net/manual/en/userlandnaming.rules.php.

> ..., also there are a few methods that don't have a return type.

I've fixed them too.

> Let me know when you feel this is ready and I can look at adding the tests

Okay, sir.

I'm currently experiencing a minor difficulty with the namespace. I'll update you, once I rounded it up.

Thank you.

solomon-ochepa:
Hi, @dcblogdev.
I'm done.

dcblogdev (approved):

PR #1822, opened 2024-04-21
[fix] fix stubs composer for app_folder

### LINKED ISSUE #1821: Custom modules.paths.app_folder value not working
image
When the `config/modules.php` value for `modules.paths.app_folder` is changed (from the default `app/` to `src/` for instance), the module will be generated using the old namespace format with the `app\` path in lowercase.

image
The module `composer.json` didn't reflect the custom app_folder, it's still using `app/`.

image
The Service Provider and every other class were created with the wrong namespace. All the namespaces have the initial `app\` segment and are in lowercase.

The custom app_folder `'src/'` wasn't generated, instead the `'app/'` is used.

Is there any workaround for this, please?
@nWidart @dcblogdev @alissn 

Thanks for the love.

alissn:
fix #1821

update stubs composer.json to read app_folder_name from config on create new module

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -54,6 +54,7 @@ return [
                 'AUTHOR_EMAIL',
                 'MODULE_NAMESPACE',
                 'PROVIDER_NAMESPACE',
+                'APP_FOLDER_NAME',
             ],
         ],
         'gitkeep' => true,
--- a/src/Commands/stubs/composer.stub
+++ b/src/Commands/stubs/composer.stub
@@ -17,7 +17,7 @@
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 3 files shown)


dcblogdev:
Noticed this does not work if the value 'APP_FOLDER_NAME', is not present in the config file.

Don't think there is a workaround to be put in place.

dcblogdev:
ah I've got a fix will for a PR for this shortly.

alissn:
@dcblogdev, thank you for fixing that. I forgot to check for this bug.🙏

I've been very busy this week and haven't had time to review GitHub.

dcblogdev:
no problem, I've got a bit more time on my hands to trying to get through the Pr's and issues a bit more.

PR #1874, opened 2024-06-20
Update provider.stub

solomon-ochepa:
Generates proper component namespace (custom namespace is fully supported).


dcblogdev:
fixed the broken tests. I'm traveling will have a proper look at this later.

PR #1879, opened 2024-06-21
Streamline Provider Registrations with a Single File

alissn:
Hi,

In this pull request, I have updated the provider registrations to use a single file.

### Changes

Currently, a file is created for each module in the `bootstrap/cache` directory. With this update, a manifest class is created to register all providers of modules in one file. I haven't deleted the old logic yet; if @dcblogdev and the community agree with this pull request, we can remove the old logic afterward.

### Next Steps

The next step is to simplify the module providers further by registering migrations, languages, components, facades, etc., inside the package to improve the package's load performance. This addresses issue #1745.

### Added Commands

1. **module:clear-compiled**
   - Deletes the compiled `modules.php` file.
   ```sh
   php artisan module:clear-compiled
   ```

2. **module:discover**
   - Recreates the compiled `modules.php` file.
   ```sh
   php artisan module:discover
   ```

### Before

```
➜  cache git:(master) ✗ tree
.
├── module0_module.php
├── module100_module.php
... // other files
├── module97_module.php
├── module98_module.php
├── module99_module.php
├── module9_module.php
├── packages.php
└── services.php
```

Each file contains:

```php
<?php return array (
  'providers' => 
  array (
    0 => 'Modules\\Module4\\Providers\\Module4ServiceProvider',
  ),
  'eager' => 
  array (
    0 => 'Modules\\Module4\\Providers\\Module4ServiceProvider',
  ),
  'deferred' => 
  array (
  ),
);
```

### After

```
➜  cache git:(master) ✗ tree
.
├── modules.php
├── packages.php
└── services.php
```

`modules.php` content:

```php
<?php return [
    'providers' => [
        0   => 'Modules\\Module0\\Providers\\Module0ServiceProvider',
        1   => 'Modules\\Module0\\Providers\\EventServiceProvider',
        ... // other providers
        152 => 'Modules\\Module99\\Providers\\Module99ServiceProvider',
    ],
    'eager' => [
        0   => 'Modules\\Module0\\Providers\\Module0ServiceProvider',
        1   => 'Modules\\Module0\\Providers\\EventServiceProvider',
        ...  // other providers
        152 => 'Modules\\Module99\\Providers\\Module99ServiceProvider',
    ],
    'deferred' => [],
];
```

If anyone has ideas to improve this pull request, please let me know.

PR #1883, opened 2024-06-30
Add Event Handling and Constants for Module Actions

alissn:
Hi,

In this pull request, I have added events for various actions and created a constant `ModuleEvent` to manage event names.

- New events added: `creating`, `created`, `deleting`, `deleted`, `used`, `unused`
- Existing events updated for usage: `boot`, `register`, `disabling`, `disabled`, `enabling`, `enabled`

This pull request depends on #1879, as provider files are registered with a single file. It's important to add a listener to clear the manifest file after certain actions.

Thanks.

### DIFF
--- a/src/Constants/ModuleEvent.php
+++ b/src/Constants/ModuleEvent.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Nwidart\Modules\Constants;
+
+class ModuleEvent
+{
+    const BOOT = 'boot';
+
+    const REGISTER = 'register';
+
+    const DISABLING = 'disabling';
+
+    const DISABLED = 'disabled';
+
+    const ENABLING = 'enabling';
+
+    const ENABLED = 'enabled';
+
+    const CREATE = 'create';
+
+    const DELETE = 'delete';
+
+    const USED = 'used';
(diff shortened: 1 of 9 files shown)


solomon-ochepa:
> Hi,
> 
> In this pull request, I have added events for various actions and created a constant `ModuleEvent` to manage event names.
> 
> * New events added: `create`, `delete`, `used`, `unused`

Nice feature, but why not use "creating" & "created" instead of "create" and "deleting" & "deleted" instead of "delete"?

Events are usually written in present continuous and past tense.

alissn:
@dcblogdev 

add new event `creating`, `created`, `deleting`, `deleted` and update tests.

dcblogdev:
Awesome, thank you!

alissn:
@dcblogdev 

Do you think it is necessary to change `boot` to `booting` and `booted`, and `register` to `registering` and `registered`?

These events may be used in projects, and changing them could cause problems.

Also, `used` and `unused` don't need to be split into two events. Do you agree?

dcblogdev:
I agree, lets not change those as you say others could be using them already.

PR #1889, opened 2024-07-04
Auto Registration of Module Migrations

### LINKED ISSUE #1881: Migration: Nothing to migrate
After the last release the migration is not working as expected, instead it goes through all the modules without finding the migrations and prints "noting to migrate" on each module

this is the [update](https://github.com/nWidart/laravel-modules/commit/6ab190dcefde58ba2ff1d1cefee39ee0c85e330c) that causes the issue.

alissn:
Hi,

In this pull request, I have added a section to enable the auto-registration of module migrations. Additionally, a config option has been added to enable or disable this feature.

After merging this pull request, you can remove the following line from the `ModuleServiceProviders` (though leaving it won't cause any issues):

```php
$this->loadMigrationsFrom(module_path($this->moduleName, 'database/migrations'));
```

> Note: This option does not register subfolders of migrations. Subfolders should be registered manually.

This pull request also fixes #1881. Apparently, some project modules were not registering their migration paths in the service provider, even though they exist in the package's stub file.

If you need more details or have any questions, please let me know.


src/LaravelModulesServiceProvider.php
```diff
@@ -74,4 +77,27 @@ protected function registerServices()
...
+        if (! $this->app['config']->get('modules.auto-discover.migrations', true)) {
+            return;
+        }
+
+        $this->app->resolving(Migrator::class, function (Migrator $migrator) {
+
+            $path = implode(DIRECTORY_SEPARATOR, [
+                $this->app['config']->get('modules.paths.modules'),
+                '*',
+                '[Dd]atabase',
```
dcblogdev, line 92: Is this correct?

alissn: The old module database directory is named `Database` with an uppercase "D", while in the new structure, it is lowercase `database`.

The `glob` function reads directories using this regex pattern.

PR #1897, opened 2024-07-21
Create Module::modules()

solomon-ochepa:
Get the list of modules:
- $enabled = null: Get all modules,
- $enabled = true: Get all enabled modules,
- $enabled = false: Get all disabled modules.


dcblogdev:
new code should be covered with tests.

Issue #949, opened 2020-03-26
Stub plural constants question


garbinmarcelo:
Hello, 
generally when we want to define stubs we use constants such as: 
- `$STUDLY_NAME$`
- `$LOWER_NAME$`
- ...

I would like to know if there are any of these constants that the name of the module, example `$LOWER_NAME$` is written in the plural.

If so, what would be the constant? If not, how would you create this? How can I do it...

Thank you for your help

nWidart:
hi,

I'm not sure what you're asking.
You can check each command for their values if you're looking for that.

garbinmarcelo:
Okay, I'll give you an example to try to explain it better.

With his package it is possible to define custom stubs, and this is one of the most amazing things I think of him.

![image](https://user-images.githubusercontent.com/9335588/77678745-60d7d800-6f70-11ea-8049-f66ea41d9f0f.png)

Usually I place the stubs folder at the root of my project and customize some files, such as the controller.stub file

![image](https://user-images.githubusercontent.com/9335588/77678849-8bc22c00-6f70-11ea-9033-966c581f2b9b.png)

In this example I'm using the constants: `$STUDLY_NAME$` and `$LOWER_NAME$` (if that's what you call it).

In this example of commented code I am using the swagger for documentation of the apis, in which case we know that the standard endpoints are used in the plural.

After running the command to create the module: `php artisan module:make Teste`
![image](https://user-images.githubusercontent.com/9335588/77679182-f8d5c180-6f70-11ea-9680-ac92af0de8ce.png)

Where I want to go is, in addition to using the constants mentioned (STUDLY, LOWER), it would be possible to use / create for example a constant called `$PLURAL_NAME$`?


In that case in the place where you have commented `/teste` would be `/testes` after creating a new module, controller, model, etc..

nWidart:
Ah, I see, thank you for the additional explanation. 😄 
These constants could be added but I'm wondering how much value these would bring.
You could also achieve this by wrapping the generation commands.

Maybe allow commands to be extended and adding new constants that way.

The easier solution would clearly be adding `_plural` on some constants, I'm only wondering if it's the best approach going forward.

garbinmarcelo:
now it’s me who don’t know if i understand very well what you mean :sweat_smile:

What you mean is to define constants in a file (without being associated with this package) and use them within other files such as: controller, model, etc ... would that be?

[...]

PR #1908, opened 2024-07-25
Module route plural paths & names

solomon-ochepa:
Based on https://github.com/nWidart/laravel-modules/issues/949

- Added two new replacements - PLURAL_LOWER_NAME and PLURAL_STUDLY_NAME
- Generate modules route path in a plural name (e.g. `/users`, `/posts`, etc.)
- Allow the route resource to auto-generate its name from the path.
- Added the a new replacement to the config file

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -39,8 +39,8 @@ return [
             'package' => 'package.json',
         ],
         'replacements' => [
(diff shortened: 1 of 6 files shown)


dcblogdev:
thanks but this will break people's projects that don't have the new key names in their config. I don't think this is needed.

solomon-ochepa:
This feature is needed, however you are right about it breaking existing projects.

I'll add a fallback key and apply for reopening.

PR #1928, opened 2024-09-01
Restructure database stubs, commands and tests

solomon-ochepa:
Proper structure will ease navigation, scalability, and maintenance of the code.

I'll restructure the remaining stubs and commands to resemble the Laravel structure for simplicity purposes if a go-ahead is given.

Thanks.


dcblogdev:
I don't think this is needed.

Issue #634, opened 2018-09-24
Module load slow


quienti:
Hi,

I have a problem loading modules.
Indeed, after having analyzed my project, I remade that the last modules are more and more slow.
I am also attaching a result of my research.

One of my last modules is a support module, with a loading order (order to 50 in module.json).
![untitled 2](https://user-images.githubusercontent.com/7326954/45956751-5803f000-bfe1-11e8-9d6e-6cb4ed0ec691.jpg)

And now, if I ask him to load first (order at 0 in module.json)
![untitled 1](https://user-images.githubusercontent.com/7326954/45956768-6225ee80-bfe1-11e8-94cc-d3c57901dcce.jpg)

We see a clear difference in loading, knowing that no code has changed.

Do you have an idea about this problem?

Thank you in advance for your help

Laravel => 5.7
Laravel Module => ^4.0
Php => 7.2.6

nWidart:
Seems you have something that's slow in your app.

Laravel modules has been tested and used with a lot of modules (10+), and still be sub 200ms.

renanBritz:
Hello!

This is a long shot but...

If you're using homestead with virtualbox, shared folders could be a major overhead. File synchronization and disk performance is just slow. Try using NFS if that's the case.

Regards!

quienti:
Hi !

No, I'm local on a windows 10 and wamp
I would have to ask my colleague what she is using because she was the one who reported the problem to me with your much higher times
(about 7 seconds)

renanBritz:
Other than that, [blackfire.io](https://blackfire.io/docs/introduction) may help you identify performance issues.

Regards!

quienti:
Okay, I have found my problem !!

````
/**
     * Register the service provider.
     *
     * @return void
     */
    public function register()
    {
        $this->registerTranslations();
        $this->app->register(RouteServiceProvider::class);
    }
````

The registerTranslations function is my problem!

But my problem is that I need to add the translations because I translate my urls in the road

An idea of ​​how I could achieve this?

quienti:
Hi,

Thank you for your help
This problem is not from this package !

Sorry !

AnthonyConklin:
[...]

PR #1921, opened 2024-08-15
Blazing Fast Module Loading 🚀😎

alissn:
Hi,

In this pull request, I've updated the logic of the `FileRepository` to significantly speed up the load time of modules by storing the result of the `scan` method in a static variable.

### Benchmarking
To test this improvement, I created 166 modules and benchmarked the `isEnabled()` method of the package.

**Benchmark Code:**
```php
    $modules = [
        "A1"        => "A1",
        "A2"        => "A2",
        "Ali"       => "Ali",
        "Auth"      => "Auth",
        "Module0"   => "Module0",
        "Module1"   => "Module1",
        "Module10"  => "Module10",
        // other modules name
        "Module98"  => "Module98",
        "Module99"  => "Module99",
        "Profile"   => "Profile",
    ];

    \Illuminate\Support\Benchmark::dd([
        'check1' => function () use ($modules) {
            $module = array_rand($modules);
            $a      = \Nwidart\Modules\Facades\Module::isEnabled($module);
        },
    ], 30);
```

**Results:**

*Before:*
image

*After:*
image

also check with [laravel-debugbar](https://github.com/barryvdh/laravel-debugbar):
*Before:*
telegram-cloud-photo-size-4-5904411011077751978-y
*After:*
telegram-cloud-photo-size-4-5904411011077751977-y

As you can see, the results are impressive! 🚀🚀🚀😎

### Additional Improvements:
I've also optimized the logic for finding modules. Previously, the module search was done inside a loop with a complexity of O(n):

```php
 foreach ($this->all() as $module) {
    if ($module->getLowerName() === strtolower($name)) {
        return $module;
    }
}
```

Now, I've changed the storage keys to lowercase when storing, and I check for module existence with an array key, reducing the complexity to O(1):

```php
   return $this->all()[strtolower($name)] ?? null;
```

This change further enhances performance and efficiency.  and can solve #1745 

Please review the changes and let me know your thoughts!

PR #1937, opened 2024-09-11
Cache modules_path

peter-mw:
I have a slow module loading when I have more than 80 modules.

Same as #634 

Adding cache to this function partially solves the problem.


peter-mw:
Its a bit hacky but here are the before-after results

Before: 2 sec

image


After: 1 sec

image

alissn:
hi, please use `dev-master` version until release new version.

on #1921 i solved this issue.

PR #1952, opened 2024-09-19
Revert Module Providers Registration to Previous Version

alissn:
Hi,

This pull request reverts the module providers registration process to the previous version. The changes were initially introduced in #1879 by me.

After reviewing the codebase and encountering new issues, it became clear that this modification introduced several breaking changes. Some projects that rely on disabling modules and registering them with a single file have experienced issues due to these updates.

Additionally, this change caused the `boot` and `register` events for modules to stop firing. This issue wasn't caught earlier because the current tests for modules do not cover this part. A separate pull request will add tests for this section to prevent similar issues in the future.


dcblogdev:
will this affect the loading of the modules that you recently introduced?

alissn:
@dcblogdev

no,  in my os, boot time not changed. has same times.

Issue #1880, opened 2024-06-23
Hot reload is not working with Vite load from modules


bee1194:
I have a module Admin. I want to load its own assets. 
These steps:
1. `{{ module_vite('build-admin', 'resources/assets/sass/app.scss') }}`
2. Run `npm run dev` 
3. Reload the page
4. Get this error `Vite manifest not found at: /var/www/html/public/build-admin/manifest.json`
After debugging, I saw Module_vite is using "storage_path('vite.hot')", but my hot file is not in storage, It is in public.
It will work when I replace `storage_path('vite.hot')` with `ViteFacade::hotFile()`

`function module_vite($module, $asset): Vite
    {
        return ViteFacade::useHotFile(ViteFacade::hotFile())->useBuildDirectory($module)->withEntryPoints([$asset]);
    }`

vincentcordel:
I got the same problem.
Your fix works and doesn't :)

It does fix the problem in the module but if you have another `npm run dev` on the root, you need to refresh it to get the update. I can't get both running at the same time.

... and my vite knowledge is too limited to understand what's going on.

stevebrainng:
I rather did this instead.

Created a new function. 

Then copied the hot file to modules build directory and I don't need to refresh on modules page or non-modules pages.



```
if (! function_exists('panel_module_vite')) {
    function panel_module_vite($module, $asset): Vite
    {
        //public_path('/build-tenant/hot')
        return ViteFacade::useHotFile(public_path($module.'/hot'))->useBuildDirectory($module)->withEntryPoints([$asset]);
    }
}
```

satrio1256:
I see that in my case adding this line under plugins.laravel works:
```
hotFile: "../../storage/vite.hot",
```

It'll be somewhat like this for me, maybe this can be adjusted so that each module can have their own hotfile?
```javascript
export default defineConfig({
  ...
  plugins: [
    laravel({
      ...
      publicDirectory: "../../public",
      hotFile: "../../storage/vite.hot"
      ...
    })
  ]
  ...
})
```

dcblogdev:
This https://github.com/nWidart/laravel-modules/pull/1965 address this issue, I'll be doing a new release shortly.

PR #1965, opened 2024-10-24
Add parameter for custom hotfile directory

### LINKED ISSUE #1906: Change vite hotfile location
Hello

Is it possible to change the hotfile location for a module?

I see in helpers it points to storage() which points to the root laravel storage path. I would like a separate hot file in my module .

I did this and it worked for me but want to know if there is another way:

```
if (! function_exists('module_vite')) {
    /**
     * support for vite
     */
    function module_vite($module, $asset): Vite
    {
        return ViteFacade::useHotFile(module_path('Retail','storage/vite.hot'))->useBuildDirectory($module)->withEntryPoints([$asset]);
    }
}

satrio1256:
Fixes #1906 #1880

This will also add possibility for the app to have multiple "hot" file simultaneously

### DIFF
--- a/src/Commands/stubs/views/master.stub
+++ b/src/Commands/stubs/views/master.stub
@@ -18,12 +18,12 @@
     <link href="https://fonts.bunny.net/css?family=figtree:400,500,600&display=swap" rel="stylesheet" />
 
     {{-- Vite CSS --}}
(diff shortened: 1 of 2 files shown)


rsd:
Please, patch 10.x series with this, for those stuck in Laravel 10 :)

PR #1991, opened 2024-11-21
Update route-provider.stub

RaihanIIUC:
as route file growing large , its hard to maintain . so i am trying to make all web.php in one umbrella's  .


dcblogdev:
I think this is better suited in your application as and when needed. This would load console and api routes into web.

PR #2010, opened 2025-01-23
fix: resolve Blade component namespace issue by removing redundant 'app' prefix in PSR-4 configuration. This caused components to not be recognized in Blade when variables were passed in the component's constructor.

reza-asgharzadeh:
fix: resolve Blade component namespace issue by removing redundant 'app' prefix in PSR-4 configuration. This caused components to not be recognized in Blade when variables were passed in the component's constructor.

The issue was caused by the redundant 'app' prefix in the PSR-4 configuration, which led to Blade components not being recognized correctly. To fix this, the 'app' prefix was removed from both the service provider and the configuration. Without this fix, the component class would not be correctly linked to Blade, and variables defined in the component's constructor would not be recognized.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -116,7 +116,7 @@ return [
             'channels' => ['path' => 'app/Broadcasting', 'generate' => false],
             'class' => ['path' => 'app/Classes', 'generate' => false],
             'command' => ['path' => 'app/Console', 'generate' => false],
-            'component-class' => ['path' => 'app/View/Components', 'generate' => false],
+            'component-class' => ['path' => 'View/Components', 'generate' => false],
             'emails' => ['path' => 'app/Emails', 'generate' => false],
             'event' => ['path' => 'app/Events', 'generate' => false],
             'enums' => ['path' => 'app/Enums', 'generate' => false],
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -109,7 +109,7 @@ class $CLASS$ extends ServiceProvider
 
         $this->loadViewsFrom(array_merge($this->getPublishableViewPaths(), [$sourcePath]), $this->nameLower);
 
-        $componentNamespace = $this->module_namespace($this->name, $this->app_path(config('modules.paths.generator.component-class.path')));
+        $componentNamespace = $this->module_namespace($this->name, config('modules.paths.generator.component-class.path'));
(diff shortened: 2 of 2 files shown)


solomon-ochepa:
Nice work. Please keep the title short nice time. You can always add a description to explain what the PR does.

config/config.php
```diff
@@ -116,7 +116,7 @@
             'channels' => ['path' => 'app/Broadcasting', 'generate' => false],
             'class' => ['path' => 'app/Classes', 'generate' => false],
             'command' => ['path' => 'app/Console', 'generate' => false],
-            'component-class' => ['path' => 'app/View/Components', 'generate' => false],
+            'component-class' => ['path' => 'View/Components', 'generate' => false],
```
solomon-ochepa, line 119: You are expected to change the value to whatever you wish in your published config file.  

This default follows Laravel's standard structure and should remain unchanged.

src/Commands/stubs/scaffold/provider.stub
```diff
@@ -109,7 +109,7 @@ class $CLASS$ extends ServiceProvider
 
         $this->loadViewsFrom(array_merge($this->getPublishableViewPaths(), [$sourcePath]), $this->nameLower);
 
-        $componentNamespace = $this->module_namespace($this->name, $this->app_path(config('modules.paths.generator.component-class.path')));
+        $componentNamespace = $this->module_namespace($this->name, config('modules.paths.generator.component-class.path'));
```
solomon-ochepa, line 112: You are expected to change the value to whatever you wish in your published config file.  

The default follows Laravel's structure, but `$this->app_path()` does not enforce `app/`. You can leave it empty or set any value through the config.

solomon-ochepa (changes requested):

solomon-ochepa:
This PR is a NO from me.

dcblogdev:
component's class's works out the box:

```bash
php artisan module:make-component Demo Profile
```

Creates:

```php
<?php

namespace Modules\Profile\View\Components;

use Illuminate\View\Component;
use Illuminate\View\View;

class Demo extends Component
{
    /**
     * Create a new component instance.
     */
    public function __construct() {}

    /**
     * Get the view/contents that represent the component.
     */
    public function render(): View|string
    {
        return view('profile::components.demo');
    }
}
```

view:

```php
<div>
  hello. I am a demo component.
</div>
```

Usage:

```php
<x-profile::demo />
```

PR #2022, opened 2025-02-14
Remove Cached Module Providers Manifest to Improve Dynamic Module Detection  

### LINKED ISSUE #2020: Module Cache Issues When Switching Branches
### Versions:
- laravel-modules Version: 11.1.8
- Laravel Version: 11.41.3
- PHP Version: 8.4.2

### Description:

Switching between branches with different modules throws an error, making it difficult to develop new modules in other branches. `php artisan module:clear-compiled` doesn't resolve this, requiring manual modules cache deletion. 

Additionally, disabled modules in `modules_statuses.json` aren't loaded after changing their status to enabled.

```
   Error 

  Class "Modules\{ModuleName}\Providers\{ModuleName}ServiceProvider" not found

  at vendor/laravel/framework/src/Illuminate/Foundation/Application.php:961
    957▕      * @return \Illuminate\Support\ServiceProvider
    958▕      */
    959▕     public function resolveProvider($provider)
    960▕     {
  ➜ 961▕         return new $provider($this);
    962▕     }
    963▕ 
    964▕     /**
    965▕      * Mark the given provider as registered.

      +11 vendor frames 

  12  [internal]:0
      Illuminate\Foundation\Application::{closure:Illuminate\Foundation\Application::boot():1127}(Object(Nwidart\Modules\LaravelModulesServiceProvider), "Nwidart\Modules\LaravelModulesServiceProvider")
      +5 vendor frames 

  18  artisan:35
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))

```

alissn:
Hi,  

In this pull request, I have removed the caching of provider data in the `bootstrap/cache/module.php` file. Storing this cache causes several issues:  

1. Changes in `module_status.json` are not detected until `module:discover` or `module:delete-compile` is executed.  
2. When copying a module directly into the `Modules` folder, Laravel does not detect it until the command is run.  
3. Switching between branches with different module statuses leads to inconsistencies (#2020).  

Since this file is only used once during Laravel's booting process, we can safely remove its caching mechanism.  

Additionally, this pull request updates the messages for `module:discover` and `module:delete-compile`, clarifying their removal in the next major version. Tests related to these commands have been marked as `@deprecated` and will be removed accordingly.  

Thanks!

fixes #2020

### DIFF
--- a/src/Commands/ModuleClearCompiledCommand.php
+++ b/src/Commands/ModuleClearCompiledCommand.php
@@ -5,6 +5,9 @@ namespace Nwidart\Modules\Commands;
 use Illuminate\Console\Command;
 use Nwidart\Modules\ModuleManifest;
 
+/**
+ * @deprecated This command is deprecated and will be removed in the next major version.
+ */
 class ModuleClearCompiledCommand extends Command
 {
     /**
@@ -27,6 +30,6 @@ class ModuleClearCompiledCommand extends Command
             @unlink($manifest->manifestPath);
         }
 
-        $this->components->info('Compiled module files removed successfully.');
+        $this->components->warn('You may stop calling the `module:clear-compiled` command.');
     }
 }
--- a/src/Commands/ModuleDiscoverCommand.php
+++ b/src/Commands/ModuleDiscoverCommand.php
@@ -5,6 +5,9 @@ namespace Nwidart\Modules\Commands;
 use Illuminate\Console\Command;
 use Nwidart\Modules\ModuleManifest;
 
+/**
+ * @deprecated This command is deprecated and will be removed in the next major version.
+ */
 class ModuleDiscoverCommand extends Command
 {
     /**
@@ -23,14 +26,6 @@ class ModuleDiscoverCommand extends Command
 
     public function handle(ModuleManifest $manifest): void
     {
-        $this->components->info('Discovering modules');
-
-        $manifest->build();
-
-        collect($manifest->providersArray())
-            ->map(fn ($provider) => preg_match('/Modules\\\\(.*?)\\\\/', $provider, $matches) ? $matches[1] : null)
-            ->unique()
-            ->each(fn ($description) => $this->components->task($description))
-            ->whenNotEmpty(fn () => $this->newLine());
+        $this->components->warn('You may stop calling the `module:discover` command.');
     }
 }
--- a/src/ModuleManifest.php
+++ b/src/ModuleManifest.php
@@ -2,7 +2,6 @@
 
 namespace Nwidart\Modules;
 
-use Exception;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Collection;
 use Nwidart\Modules\Contracts\ActivatorInterface;
@@ -12,7 +11,7 @@ class ModuleManifest
     /**
      * The filesystem instance.
      *
-     * @var \Illuminate\Filesystem\Filesystem
+     * @var Filesystem
      */
     private $files;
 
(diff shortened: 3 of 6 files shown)


dcblogdev:
@alissn could you have a look at these conflicts, cheers.

src/Commands/ModuleDiscoverCommand.php
```diff
@@ -23,14 +26,6 @@ class ModuleDiscoverCommand extends Command
...
-        $this->components->info('Discovering modules');
-
-        $manifest->build();
-
-        collect($manifest->providersArray())
-            ->map(fn ($provider) => preg_match('/Modules\\\\(.*?)\\\\/', $provider, $matches) ? $matches[1] : null)
-            ->unique()
-            ->each(fn ($description) => $this->components->task($description))
-            ->whenNotEmpty(fn () => $this->newLine());
+        $this->components->warn('You may stop calling the `module:discover` command.');
```
solomon-ochepa, line 29: Could you remove the entire method instead? If it does nothing, then it becomes a useless code.

src/Commands/ModuleClearCompiledCommand.php
```diff
@@ -27,6 +30,6 @@ public function handle(ModuleManifest $manifest): void
             @unlink($manifest->manifestPath);
         }
 
-        $this->components->info('Compiled module files removed successfully.');
+        $this->components->warn('You may stop calling the `module:clear-compiled` command.');
```
solomon-ochepa, line 33: It would be better to tag the message, instead of replacing it while the method still operates.
```
$this->components->info('[deprecated] Compiled module files removed successfully.');
```

solomon-ochepa (changes requested):

src/ModuleManifest.php
```diff
@@ -12,7 +11,7 @@ class ModuleManifest
     /**
      * The filesystem instance.
      *
-     * @var \Illuminate\Filesystem\Filesystem
+     * @var Filesystem
```
solomon-ochepa, line 14: Nice work in making the code clean and readable!

Could you use strict data type casting instead if you are certain?
```
private Filesystem $files;
```

Using this method, there would be fewer bugs as the property will prevent unsupported data types.

solomon-ochepa (changes requested):

src/Commands/ModuleDiscoverCommand.php
alissn: are you read this line? i'll delete this file in version 12.

```
/**
 * @deprecated This command is deprecated and will be removed in the next major version.
 */
 ```

File: src/Commands/ModuleClearCompiledCommand.php (hunk 2 above)
Comment (alissn) [reply src/Commands/ModuleClearCompiledCommand.php:33]: are you read this line? i'll delete this file in version 12.

```
/**
 * @deprecated This command is deprecated and will be removed in the next major version.
 */
 ```
 
 you can see this message on [laravel horizon command](https://github.com/laravel/horizon/blob/5.x/src/Console/PublishCommand.php)

alissn:
@dcblogdev 

Conflicts have been resolved. Let me know if anything else is needed. Cheers!

src/Commands/ModuleDiscoverCommand.php
solomon-ochepa: > are you read this line? i'll delete this file in version 12.
> 
> ```
> /**
>  * @deprecated This command is deprecated and will be removed in the next major version.
>  */
> ```

Yes, I did read the documentation.

When a method is deprecated, it should either retain its logic for backward compatibility or forward calls to its replacement. In your case, you removed the logic but kept the method, leaving it as unused code.

Again, instead of "You may stop calling the `module:discover` command.", why not something more polite and professional like "The `module:discover` command has been deprecated."

src/Commands/ModuleClearCompiledCommand.php
solomon-ochepa: > are you read this line? i'll delete this file in version 12.
> 
> ```
> /**
>  * @deprecated This command is deprecated and will be removed in the next major version.
>  */
> ```
> 
> you can see this message on [laravel horizon command](https://github.com/laravel/horizon/blob/5.x/src/Console/PublishCommand.php)

The documentation message is very okay. I was referring to the `$this->components->warn()` message.

I understood it would be removed subsequently, but while the code exists, it should still pass the correct warning/error message.

Thank you.

PR #2030, opened 2025-02-25
Added missing type hinting for helpers.php

edgarsn:
Upgrading to larastan 3.0, caused a lot of phpstan errors related to `module_path()` usage. 

PHPStan assumes `module_path()` returns mixed type without any typehints, causing coode below to fail:

```php
<?php

if (is_dir(module_path('my_module', 'views'))) { // Parameter #1 $filename of function is_dir expects string, mixed given. 
   // ....
}
```

Also I added parameter type hints for `module_vite()` as well.

To not introduce potential breaking change using native types, I sticked to phpdoc syntax.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -4,6 +4,11 @@ use Illuminate\Foundation\Vite;
 use Illuminate\Support\Facades\Vite as ViteFacade;
 
 if (! function_exists('module_path')) {
+    /**
+     * @param string $name
+     * @param string $path
+     * @return string
+     */
     function module_path($name, $path = '')
     {
         $module = app('modules')->find($name);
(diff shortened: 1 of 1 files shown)


dcblogdev (approved):

edgarsn:
Sorry about the laravel pint fixes mess. Last commit should fix it.

dcblogdev:
Thanks for sorting this

solomon-ochepa:
Nice, thanks for that fix. But I think the strict type casting would be better;
```
if (! function_exists('module_path')) {
    function module_path(string $name, string $path = ''): string
    {
        $module = app('modules')->find($name);
        return $module->getPath().($path ? DIRECTORY_SEPARATOR.$path : $path);
    }
}
```

And

```
if (! function_exists('module_vite')) {
    /**
     * support for vite
     */
    function module_vite(string $module, string $asset, ?string $hotFilePath = null): Vite
    {
        return ViteFacade::useHotFile($hotFilePath ?: storage_path('vite.hot'))->useBuildDirectory($module)->withEntryPoints([$asset]);
    }
}
```

This will cover both the Docblock and strict type casting.

Thanks a lot.

dcblogdev (approved):

PR #2036, opened 2025-03-06
[new feature] Module plural names

solomon-ochepa:
Based on: https://github.com/nWidart/laravel-modules/issues/949
Ref: https://github.com/nWidart/laravel-modules/pull/1908

Added two new replacements: PLURAL_LOWER_NAME and PLURAL_STUDLY_NAME

This feature is necessary for many functionalities, such as the route path and name.

This will no longer break any existing setup as it's not implemented in any stub. Dev are expected to implement it at their choice.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -39,8 +39,8 @@ return [
             'package' => 'package.json',
         ],
         'replacements' => [
-            'routes/web' => ['LOWER_NAME', 'STUDLY_NAME', 'KEBAB_NAME', 'MODULE_NAMESPACE', 'CONTROLLER_NAMESPACE'],
-            'routes/api' => ['LOWER_NAME', 'STUDLY_NAME', 'KEBAB_NAME', 'MODULE_NAMESPACE', 'CONTROLLER_NAMESPACE'],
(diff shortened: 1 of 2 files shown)


dcblogdev (approved):

fabioselau077:
Let me understand: It's not a rule, but is it recommended to create the module in the plugin or singular? Would it be, for example, User or Users? Post or Posts? 

In doc says `php artisan module:make posts`

fabioselau077:
> > Let me understand: It's not a rule, but is it recommended to create the module in the plugin or singular? Would it be, for example, User or Users? Post or Posts?
> > In doc says `php artisan module:make posts`
> 
> * `User`, not `Users`,
> * `Post` not `Posts`.
> 
> However, it's recommended to use plural for the routes, which is what this PR was created for.
> 
> ### Example:
> Module: `User` Route: `users.index`, `users.create`, `users.store`, `users.show`, etc. Model: `User` (`modules/User/app/Models/User.php` ~ namespace: `Modules\User\App\Models\User`) Table: `users` Controller: `UserController.php`
> 
> The example on the official documentation is incorrect and will be updated soon.
> 
> Thanks.

Got it. I was unsure, so I found another issue and then this PR, but the documentation makes it even more confusing.

In my case, it'll just be the API (--api, no resources, routes/web.php)... so I can create the module and everything in the singular, as `User`, right? Except, of course, the database table, which will be `users`.

Thx for reply!

PR #2053, opened 2025-04-21
Allow Module Configs to Override Root Application Configs

solomon-ochepa:
#### Summary:
This PR introduces a core enhancement to the `registerConfig()` logic. It ensures that configuration files within a module (e.g., `modules/Blog/config/config.php` or `modules/Blog/config/blog.php`) override Laravel's root application configs (e.g., `config/blog.php`) during registration.

#### Why This is Necessary:
The modular architecture encourages encapsulation of logic and behavior per domain (module). However, Laravel’s default `mergeConfigFrom()` merges shallowly and always prioritizes existing config values, meaning module-defined values are ignored if a root config already exists.

This directly contradicts the core idea of modularity - that a module should define and control its behavior without being limited by app-level configs.

By allowing module configs to override root configs, developers get:
- Predictable behavior across all modules.
- Clean overrides and centralization of domain logic.
- Better separation of concerns between global app config and module logic.

#### Before:
```php
// config/blog.php
return [
    'comments' => false,
];

// Modules/Blog/config/blog.php
return [
    'comments' => true,
];

// config('blog.comments') === false
```

#### After:
```php
// config('blog.comments') === true
```

Module config values now correctly override root config values.

### DIFF
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -91,10 +91,21 @@ class $CLASS$ extends ServiceProvider
                     $key = ($relativePath === 'config.php') ? $this->nameLower : $configKey;
 
                     $this->publishes([$file->getPathname() => config_path($relativePath)], 'config');
-                    $this->mergeConfigFrom($file->getPathname(), $key);
+                    $this->merge_config_from($file->getPathname(), $key);
                 }
             }
         }
+    }
+
+    /**
+     * Merge config from the given path recursively.
+     */
+    protected function merge_config_from(string $path, string $key): void
+    {
+        $existing = config($key, []);
+        $module_config = require $path;
+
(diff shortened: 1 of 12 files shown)


rsd:
Why is this necessary?

I am working on a deep merge of configuration keeping the top-level/root config values.

I believe this should work on the other way around.
A module's config should have default values.  These default values can be final if the top-level config does not exists (no clash).
However, if after installing a module, you do not agree that these config values in the module are the correct one for you app, you replace them at the application top-level, without the need to edit the module's config file.

I am clearly missing some (maybe obvious) point.
Could you please elaborate it to me?

BTW, my plan is to have the other way around behavior as an alternative and not as a replacement of the current take.

thanks,
-rsd

solomon-ochepa:
The shortest answer is, a module extends the behavior of the root app. Extends!
________________________________

PR #1735, opened 2024-02-05
Implement base command

alissn:
Hi,

In this pull request, I've begun refactoring the structure of the command. The rules for all commands are as follows:

1. When called without any arguments, it will show a multiselect of Laravel prompts to choose which module should be checked.
2. When the option `--all` is passed, it will check all modules.
3. Module names can be passed as arguments to the command.

Example:

### Option 1:
   ```sh
   php artisan module:use
   ```
image

---

### Option 2:
   ```sh
   php artisan module:use --all
   ```
image

---

### Option 3:
   ```sh
   php artisan module:use base wallet
   ```
image

In this PR, I've updated the following commands:
- [x] module:use
- [x] module:unuse
- [x] module:update
- [x] module:publish
- [x] module:publish-config
- [x] module:publish-migration
- [x] module:publish-translation
- [x] module:delete
- [x] module:migrate
- [x] module:migrate-status
- [x] module:migrate-rollback
- [x] module:migrate-reset
- [x] module:migrate-refresh
- [x] module:disable
- [x] module:enable
- [x] module:dump
- [x] module:lang
- [x] module:seed

PR #2074, opened 2025-05-18
Support 'All' as argument in module:seed command to seed all modules

itsalifadaei:
Previously, the module:seed command required passing each module name individually, or calling the command with no argument (and optionally typing 'all' when prompted).

With this change, you can now do:

php artisan module:seed All

…and it will correctly seed all registered modules.

This improves usability and allows scripting tools or automated workflows to explicitly call seeding for all modules via a single consistent command.


alissn:
Hi,
the `--all` option already exists. try this command

```bash
php artisan module:seed --all
```

more info #1735

PR #2089, opened 2025-06-20
Ensure modules are registered in register phase

tareq-alqadi:
Moved module registration to the `register()` method to ensure that module service providers are loaded early and take their proper place in the Laravel lifecycle. Previously, module service providers were being registered during the `boot()` phase, which caused their `register()` methods to run too late — during the boot stage — leading to incorrect and unintended behavior.

Additionally, removed the cacheManager dependency from the FileRepository, as it is not available in the `register()` phase and was not being used.

### DIFF
--- a/src/FileRepository.php
+++ b/src/FileRepository.php
@@ -3,7 +3,6 @@
 namespace Nwidart\Modules;
 
 use Countable;
-use Illuminate\Cache\CacheManager;
 use Illuminate\Container\Container;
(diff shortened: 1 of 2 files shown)


dcblogdev:
@alissn can I get your opinion on this PR

alissn:
Hi,
@dcblogdev That sounds logical. Moving the module registration to the `register` method is a good approach.

thanks @tareq-alqadi

PR #2101, opened 2025-08-09
[fix] fix return type of method

### LINKED ISSUE #2096: Nwidart\Modules\Support\Migrations\NameParser::getTableName(): Return value must be of type string, null returned
### Versions:
- laravel-modules Version: 12.0
- Laravel Version: 12.0
- PHP Version: 8.2

### Description:

When try to run this command 
`php artisan module:make-migration update_users_table_add_admin_group_id System` 

it gives following error :

```
   TypeError

  Nwidart\Modules\Support\Migrations\NameParser::getTableName(): Return value must be of type string, null returned

  at vendor/nwidart/laravel-modules/src/Support/Migrations/NameParser.php:73
     69▕     public function getTableName(): string
     70▕     {
     71▕         $matches = array_reverse($this->getMatches());
     72▕
  ➜  73▕         return array_shift($matches);
     74▕     }
     75▕
     76▕     /**
     77▕      * Get matches data from regex.

      +16 vendor frames

  17  artisan:16
      Illuminate\Foundation\Application::handleCommand(Object(Symfony\Component\Console\Input\ArgvInput))
```

### Steps To Reproduce:

Run any command like `php artisan module:make-migration update_users_table_add_admin_group_id System`
You will get the error

To avoid above error has to run it as : 
`php artisan module:make-migration add_admin_group_id_to_users_table System` 

But if the table name could not extract from the command, still migration should be created without the table name.

alissn:
Hi,

This pull request fixes #2096.
The `getTableName` method’s return type can be null, so its type declaration has been updated.

### DIFF
--- a/src/Support/Migrations/NameParser.php
+++ b/src/Support/Migrations/NameParser.php
@@ -66,7 +66,7 @@ class NameParser
     /**
      * Get the table will be used.
      */
-    public function getTableName(): string
+    public function getTableName(): ?string
     {
         $matches = array_reverse($this->getMatches());
 


alissn:
@solomon-ochepa

I’m sure that changing without adding tests can cause new issues.
Your latest pull request seems to follow this pattern.

Currently, this package has only **64% test coverage**, so we can’t be sure all the changes you made are tested and will work fine.

alissn:
@solomon-ochepa

Rather than only updating the stub file and renaming methods, it would be more valuable to contribute by adding additional test coverage or developing new features, as this would have a greater positive impact on the project.

This PR doesn’t require tests. Refactoring this part would be better, but I don’t have the time right now. If you can, please create a new PR for that.

solomon-ochepa:
> This PR doesn’t require tests. Refactoring this part would be better, but I don’t have the time right now. If you can, please create a new PR for that.

## Suggested Tests for getTableName(): ?string
1. Returns correct string when matches exist
 - Setup a scenario where getMatches() returns a non-empty array (e.g., ['foo', 'bar', 'baz']).
 - Ensure getTableName() returns the first element after reversing—i.e., 'foo'.

2. Returns null when the matches array is empty
 - Mock getMatches() to return an empty array ([]).
 - Confirm that getTableName() returns null.

3. Correct behavior when matches only contain one element
 - getMatches() returns something like ['only'].
 - Expect getTableName() to return 'only'.

4. Behavior with null values inside matches
 - If getMatches() returns arrays including nulls, e.g., ['one', null, 'two'].
 - Ensure getTableName() handles it gracefully (i.e., returns whatever array_shift gives—likely 'one', but test that explicitly).

5. Access visibility & method chaining
 - Test that getTableName() is publicly accessible and works via the class interface without errors.

alissn:
@solomon-ochepa

For more information, please check the linked issue and run the tests on your local environment. If you have suggestions for additional tests, feel free to create a new pull request.

I’m not interested in arguing, but I’m happy to answer questions from anyone except you.

PR #2114, opened 2025-09-16
Refactor BaseCommand to improve module resolution and add support for…

solomon-ochepa:
### Summary
Currently, when running migrations, modules are loaded in ascending (asc) alphabetical order, which causes issues with database tables' foreign key constraints.

For example, suppose an `Entry` module depends on a `Transaction` module, but the Entry module is processed first due to alphabetical ordering. In that case, the migration will fail because the required foreign key (transaction_id) references don't exist yet.

### Describe in detail what you propose, show (preferable) code examples and also signal if you're willing to work on it!
```php
    /**
     * Read the statuses file defined in config('modules.activators.file.statuses-file')
     * and return enabled module names preserving the order in that file.
     * Falls back to repository ordering if file is missing or invalid.
     */
    private function getModulesOrderedByStatusesFile(): array
    {
        $statusesFile = $this->laravel['config']->get('modules.activators.file.statuses-file');
        $statusesFile = is_string($statusesFile) && $statusesFile !== ''
            ? $statusesFile
            : base_path('modules_statuses.json');

        $sort = $this->option('direction') ?? 'asc';

        if (! is_string($statusesFile) || ! file_exists($statusesFile)) {
            return array_keys($this->laravel['modules']->getOrdered($sort));
        }

        $json = @file_get_contents($statusesFile);
        $data = is_string($json) ? json_decode($json, true) : null;

        if (! is_array($data)) {
            return array_keys($this->laravel['modules']->getOrdered($sort));
        }

        $modules = [];
        foreach ($data as $name => $enabled) {
            if ($enabled === true || $enabled === 1) {
                $modules[] = $name; // Keep StudlyCase names as defined
            }
        }

        // Fallback if no enabled modules found
        if (empty($modules)) {
            return array_keys($this->laravel['modules']->getOrdered($sort));
        }

        return $modules;
    }
```

@dcblogdev

### DIFF
--- a/src/Commands/BaseCommand.php
+++ b/src/Commands/BaseCommand.php
@@ -8,6 +8,7 @@ use Illuminate\Console\Prohibitable;
 use Illuminate\Contracts\Console\PromptsForMissingInput;
 use Illuminate\Support\Collection;
 use Nwidart\Modules\Contracts\ConfirmableCommand;
+use Nwidart\Modules\Module;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


src/Commands/BaseCommand.php
```diff
@@ -129,13 +132,53 @@ protected function promptForMissingArguments(InputInterface $input, OutputInterf
...
+        foreach ($data as $name => $enabled) {
+            if ($enabled === true || $enabled === 1) {
+                $modules[] = $name; // Keep StudlyCase names as defined
+            }
+        }
+
+        // Fallback if no enabled modules found
+        if (empty($modules)) {
+            return array_keys($this->laravel['modules']->getOrdered($sort));
+        }
```
alissn, line 177: This part is incorrect. If no modules are enabled, the command shouldn’t run across all modules. Running migrations when no modules are enabled doesn’t make sense and could cause unintended side effects.

PR #2142, opened 2026-03-10
Fix module config overwriting cached env values with null after config:cache

### LINKED ISSUE #2126: Config cache gets overwritten to null by provider stub (custom merge) when using env()
### Versions:
- laravel-modules Version: 12.0.4
- Laravel Version: 12.34.0
- PHP Version: 8.4.10

### Description:

The default service provider stub generated by php artisan module:make merges module config at runtime using a custom merge_config_from() inside boot(). When the application config is cached (php artisan config:cache), this code path overwrites the baked values with null for any keys that come from env().

Replacing the custom merge with $this->mergeConfigFrom() (in register()) resolves the issue.

### Steps To Reproduce:

1 Create a new Laravel project: ```composer create-project laravel/laravel test-project```
2 Require this package: ```composer require nwidart/laravel-modules```
3 Generate a module: ```php artisan module:make TestModule```
4 In ```Modules/TestModule/config/config.php``` add a test env backed key:
```
[
    'name'    => 'TestModule',
    'env_test' => env('ENV_TEST'),
]
```
5 Add the corresponding value to .env: ``` ENV_TEST='test' ```
6 Build config cache: 
```
php artisan optimize:clear
php artisan config:cache
```
7 Inspect using Tinker:
```
php artisan tinker
>>> app()->configurationIsCached();  // true
>>> config('testmodule');
/*
Expected: values from .env
Actual:   ["name" => "TestModule", "env_test" => null]
*/
```
8 Compare with the cached file:
```
>>> (require base_path('bootstrap/cache/config.php'))['testmodule'];
/*
Shows the correct baked values:
['name' => 'TestModule', 'env_test' => 'test']
*/
```

### Possible cause
The stub’s boot() calls registerConfig() which requires each module config file (require $path) and then calls config([$key => array_replace_recursive($existing, $module_config)]).
When config is cached, Laravel does not read .env at runtime. env('...') inside a required config file returns null.
The custom merge runs after the cached config is loaded and replaces previously correct values with these nulls (module array takes precedence over existing config).
Commenting out the `$this->registerConfig();` call inside `TestModuleServiceProvider::boot()`, placing
```
$this->mergeConfigFrom(
    __DIR__ . '/../../config/config.php', 'testmodule'
);
```
inside `TestModuleServiceProvider::register()`, reloading the code and rebuilding the cache:
```
php artisan optimize:clear
composer dump-autoload -o
php artisan config:cache  
```
restores the expected behaviour:
```
php artisan tinker
>>> config('testmodule');
/*
Shows the correct baked values:
['name' => 'TestModule', 'env_test' => 'test']
*/
```

isaackaara:
Fixes #2126

## What broke

After running `php artisan config:cache`, module config keys backed by `env()` come back as `null` instead of their real values:

```php
config('testmodule.env_test'); // null (should be 'test')
```

The cached config file itself is correct — the problem only surfaces at runtime.

## Root cause

The provider stub's `merge_config_from()` method is called inside `boot()` on every request. It does:

```php
$module_config = require $path;  // re-executes the config file
config([$key => array_replace_recursive($existing, $module_config)]);
```

When the config is cached, Laravel doesn't load `.env` at runtime, so all `env()` calls inside the required config file return `null`. Those nulls are then merged over the correctly baked cached values, clobbering them.

## Fix

Skip the merge when configuration is already cached. The cached config already has the correct baked-in values — re-requiring and re-merging the config file at runtime is both unnecessary and destructive.

```php
protected function merge_config_from(string $path, string $key): void
{
    if (app()->configurationIsCached()) {
        return;
    }

    // ... existing merge logic
}
```

Snapshot tests updated to match.

### DIFF
--- a/src/Commands/stubs/scaffold/provider.stub
+++ b/src/Commands/stubs/scaffold/provider.stub
@@ -111,6 +111,10 @@ class $CLASS$ extends ServiceProvider
      */
     protected function merge_config_from(string $path, string $key): void
     {
+        if (app()->configurationIsCached()) {
+            return;
+        }
+
         $existing = config($key, []);
         $module_config = require $path;
 
--- a/tests/Commands/Make/__snapshots__/ModuleMakeCommandTest__test_it_generate_module_when_provider_is_enable_and_route_provider_is_disable__1.txt
+++ b/tests/Commands/Make/__snapshots__/ModuleMakeCommandTest__test_it_generate_module_when_provider_is_enable_and_route_provider_is_disable__1.txt
@@ -111,6 +111,10 @@ class BlogServiceProvider extends ServiceProvider
      */
     protected function merge_config_from(string $path, string $key): void
     {
+        if (app()->configurationIsCached()) {
+            return;
+        }
+
         $existing = config($key, []);
         $module_config = require $path;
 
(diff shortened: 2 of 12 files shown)


dcblogdev:
thank you, I've added this to the new ModuleServiceProvider

